PR #2, opened 2022-04-29
Instructions for running via Docker

ajtrichards:
I thought i'd add some instructions for getting up and running with the `Dockerfile` that's in the repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,17 @@ yarn start
 
 Go to https://localhost:5006 in your browser and you'll see Actual.
 
+## Running via Docker
+To run using a Docker container you can use following commands;
+
+```
+git clone https://github.com/actualbudget/actual-server.git
+cd actual-server
+docker build -t actual-server .
+docker run -p 5006:5006 actual-server
+```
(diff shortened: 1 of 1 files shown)


brtwrst:
You might want to include a mount of the /data dicrectory to enable data persistance

jlongster:
Good call about `/data` -- how would that look here? Can you update it?

avioli:
Usually something like:

```sh
mkdir ./data
docker run -p 5006:5006 -v "$(pwd)/data":/data actual-server
```

Depending on throughput above method might have certain slowdowns, since it will be transferring data over the bridge to the host's file system.

An alternative would be to use `docker volume` as described here: https://docs.docker.com/storage/volumes/#create-and-manage-volumes and use `docker run ... --mount source=myvolume,target=/data ...`

Backing up the latter is described on the same page as above: https://docs.docker.com/storage/volumes/#backup-restore-or-migrate-data-volumes

jlongster:
Will merge this for now. We can iterate!

PR #6, opened 2022-04-29
Add hostname binding

UnexomWid:
Added an option in the config file to bind to a given hostname (e.g. `localhost`). I intend to run the server behind a reverse proxy and I want to bind it localhost, so I thought others may also use this option :)

I've left the default to `0.0.0.0` in order to not break the current behavior.


jlongster:
Sounds good, thanks!

PR #7, opened 2022-04-30
added a basic docker-compose file

Kk-ships:


nathanjisaac (commented):
Is `/data` the correct volume location for the actual server data?

matthewbahr:
Isn't it better to put the data in /usr/share/?

That's what my homeassistant docker compose does.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+services:
+  actual_server:
+    container_name: actual_server
+    build:
+      context: ./
+      dockerfile: Dockerfile
+    ports:
+      - "5006:5006"
+    volumes:
+      - ./data:/data
```
nathanjisaac, line 11:
```suggestion
      - ./server-files:/app/server-files
      - ./user-files:/app/user-files
```

I think this is what you want for the volumes. 

I'm matches these in the `Dockerfile`. 

```Dockerfile
RUN mkdir ./server-files
RUN mkdir ./user-files
```

nathanjisaac:
I just tested this locally on my machine. Seems to work good.

fedme (approved):
Looks good to me! (I am not a maintainer, just a random dev :smiley:)

jlongster:
Is this still up-to-date given the latest docker file changes?

Kk-ships:
@jlongster docker-compose file would be Dockerfile agnostic. I just builds a docker container based on Dockerfile contents.

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: "3"
+services:
+  actual_server:
+    container_name: actual_server
+    build:
+      context: ./
+      dockerfile: Dockerfile
```
nathanjisaac, line 7: @jlongster It's dynamically referenced here.

jlongster:
Yes but now we have two different docker files after #22, `Dockerfile` and `Dockerfile.alpine`. Does that make a difference?

jlongster:
Thanks, I'll add  this!

PR #19, opened 2022-05-01
Add GH Action workflow to publish Docker image

m3nu:
This adds a simple Github workflow to publish an official Docker image. I usually default to Github's registry to keep everything in one place. Of course the registry would be easy to adjust.

I usually publish a new image for tagged pushes, but this can be adjusted too.


m3nu:
Oh, already done. Never mind then.

PR #21, opened 2022-05-01
Health Check Endpoint

Silvenga:
I run everything on my K8s cluster, so a health check endpoint would be optimal. The container could make use of this health check too.

This pr adds the simple express-actuator middleware, which provides:

```
/health
{
  "status": "UP"
}
```

```
/info
{
  "build": {
    "name": "actual-sync",
    "description": "actual syncing server",
    "version": "1.0.0"
  }
}
```

```
/metrics
{
  "mem": {
    "rss": 45621248,
    "heapTotal": 22167552,
    "heapUsed": 20089648,
    "external": 949596,
    "arrayBuffers": 45238
  },
  "uptime": 349.9881051
}
```

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,5 +1,6 @@
 const fs = require('fs');
 const express = require('express');
+const actuator = require('express-actuator');
 const bodyParser = require('body-parser');
 const cors = require('cors');
 const config = require('./load-config');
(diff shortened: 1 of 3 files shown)


jlongster:
Nice, this sounds good.

Eventually, I'd like to figure out how to make the server more extensible, letting you "plugin" small apps to add whatever endpoints you want. But for now, I'll merge this.

piyushpaliwal:
Thanks for this PR, just wanted to see if there is any example of utilizing it within the docker-compose file. As I understand, the `curl` command is not available within the image so not sure how are we expected to use it.

j-f1:
Can you open an issue and attach what you have so far @piyushpaliwal? This PR is pretty old.

Silvenga:
@piyushpaliwal This won't work for Docker Compose files - this is only for generic health checks used by load balancers, or in my exact use-case, Kubernetes.

PR #26, opened 2022-05-04
Added easy directable url in actual budget

fzn0x:
Use the `open` package and modify the hostname from 0.0.0.0 to localhost to allow open default browser by default.

The projects are used to filter docker and anything else so there is no memory leak possibility if you run the server on a non-desktop environment. Reference: https://github.com/sindresorhus/open/blob/main/index.js.

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const open = require('open');
 const express = require('express');
 const actuator = require('express-actuator');
 const bodyParser = require('body-parser');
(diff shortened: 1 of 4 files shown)


fzn0x:
Thanks for the amazing work you put in and make this open source!

jlongster-stripe:
Thanks! I don't love these change tbh; I don't love servers that pop open a browser automatically. Half the time it gets in the way and is distracting when I'm trying to multi-task. Also we explicitly changed it from `localhost` to `0.0.0.0` so it works on all ips in another PR.

I'm going to close this, but if a lot of other people want it we can reconsider. Want to keep things simple if we can.

jlongster:
Oops, posted from my work account!

fzn0x:
> Oops, posted from my work account!

Haha, thanks for your respond I appreciate it

PR #28, opened 2022-05-04
Transition to typescript

PartyLich:
- adds basic typescript support. scripts for typechecking and compilation
- updated dockerfiles 
    - include a build step
    - best practices (with references in the commits). 
    - The RUN consolidation might need discussion, as it could have an effect on caching that you find undesirable
- app-plaid.js seems to be unused. there's no plaid dependency in the the package, and I didnt spot any references to app-plaid. I haven't removed it, but it seems like a candidate for removal
  - util/handle-error also unused
  - sync-simple also unused
- No files have been rewritten to add typings or other TS specific changes in this PR. One fix was made as a result of running the typechecker.
  - initial file conversion and very rough typings (ie there are still `any`s, which is unpleasant) in [a separate branch](https://github.com/PartyLich/actual-server/tree/refactor/typescript-conversion). At least one typing looks like it should be defined/is dependent on @actual-app/api
- clean script for build artifacts currently uses `rm -rf`, which isn't portable as far as i know. rimraf and other solutions may be portable, but I didnt want to add those without discussion

I'm assuming this is still part of the plan based on [this](https://github.com/actualbudget/actual/blob/master/CONTRIBUTING.md#transitioning-to-typescript). Both docker builds have been updated and seem to be working in my limited testing. 

Running locally without docker feels a little convoluted as I currently have it set up. There's only one extra command, but atm I'm copying node_modules and that feels...odd. Including something like [ts-node](https://typestrong.org/ts-node) would let us JIT compile any typescript and avoid the current inconvenience, but I didn't want to introduce that dependency (even if it's only for local runs) without some sort of consensus.

Adding typescript-eslint would also make sense, though I have no idea what specific rules you'd like to enforce. I can set it up with the defaults as a starting point, if desired ([done in this branch](https://github.com/PartyLich/actual-server/tree/chore/add-ts-linting)).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ bin/large-sync-data.txt
 user-files
 server-files
 fly.toml
+build/
--- a/load-config.js
+++ b/load-config.js
@@ -1,5 +1,6 @@
 let config;
 try {
+  // @ts-expect-error TS2307: we expect this file may not exist
   config = require('./config');
 } catch (e) {
   let fs = require('fs');
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "start": "node app",
     "lint": "eslint .",
+    "build": "tsc",
     "types": "tsc --noEmit --incremental",
     "verify": "yarn -s lint && yarn types"
   },
--- a/package.json
+++ b/package.json
@@ -6,7 +6,9 @@
   "main": "index.js",
   "scripts": {
     "start": "node app",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "types": "tsc --noEmit --incremental",
+    "verify": "yarn -s lint && yarn types"
   },
   "dependencies": {
     "@actual-app/api": "^4.0.1",
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "eslint": "^8.15.0",
     "eslint-plugin-prettier": "^4.0.0",
     "prettier": "^2.6.2",
+    "@types/better-sqlite3": "^7.5.0",
     "@types/node": "^17.0.31",
     "typescript": "^4.6.4"
   }
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "eslint": "^8.15.0",
     "eslint-plugin-prettier": "^4.0.0",
     "prettier": "^2.6.2",
+    "@types/node": "^17.0.31",
     "typescript": "^4.6.4"
   }
 }
--- a/sync-simple.js
+++ b/sync-simple.js
@@ -75,7 +75,6 @@ function sync(messages, since, fileId) {
          WHERE timestamp > ?
          ORDER BY timestamp`,
     [since],
-    true
   );
 
   let trie = addMessages(db, messages);
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -16,5 +16,5 @@
     "moduleResolution": "Node",
     "module": "ESNext"
   },
-  "exclude": ["node_modules", "build"]
+  "exclude": ["node_modules", "build", "./app-plaid.js"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,6 +4,7 @@
     // DOM for URL global in Node 16+
     "lib": ["ES2021", "DOM"],
     "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
     "experimentalDecorators": true,
     "resolveJsonModule": true,
     "downlevelIteration": true,
@@ -11,10 +12,9 @@
     // Check JS files too
     "allowJs": true,
     "checkJs": true,
-    // Used for temp builds
-    "outDir": "build",
-    "moduleResolution": "Node",
-    "module": "ESNext"
+    "moduleResolution": "node",
+    "module": "commonjs",
+    "outDir": "build"
   },
   "exclude": ["node_modules", "build", "./app-plaid.js"]
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -86,6 +86,11 @@
   resolved "https://registry.yarnpkg.com/@types/json-schema/-/json-schema-7.0.11.tgz#d421b6c527a3037f7c84433fd2c4229e016863d3"
   integrity sha512-wOuvG1SN4Us4rez+tylwwwCV1psiNVOkJeM3AUWUNWg/jDQY2+HE/444y5gc+jBmRqASOm2Oeh5c1axHobwRKQ==
 
+"@types/node@^17.0.31":
+  version "17.0.31"
+  resolved "https://registry.yarnpkg.com/@types/node/-/node-17.0.31.tgz#a5bb84ecfa27eec5e1c802c6bbf8139bdb163a5d"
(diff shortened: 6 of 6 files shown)


willyp713:
PLAID is intended to be used with their free individual dev account to sync directly with banks for automatic import, per the creator's blog post announcing open source. I believe nothing has actually been done to hook this up, though.

willyp713:
@PartyLich the section detailing future support for Plaid is on this page: https://actualbudget.com/open-source

Although as you point out - it's just an idea at this point, apparently without any implementation.

TomAFrench:
> initial file conversion and very rough typings (ie there are still anys, which is unpleasant) in [a separate branch](https://github.com/PartyLich/actual-server/tree/refactor/typescript-conversion). At least one typing looks like it should be defined/is dependent on @actual-app/api

@PartyLich I managed to miss this bullet point and ended up duplicating this work. I've rebased my migration branch on top of yours which also applies a number of linting fixes and typing improvements. Are you open to a PR onto your fork so we can merge these together?

> Adding typescript-eslint would also make sense, though I have no idea what specific rules you'd like to enforce. I can set it up with the defaults as a starting point, if desired ([done in this branch](https://github.com/PartyLich/actual-server/tree/chore/add-ts-linting)).

#33 

:man_facepalming: I need to read more carefully.

PartyLich:
> > initial file conversion and very rough typings (ie there are still anys, which is unpleasant) in [a separate branch](https://github.com/PartyLich/actual-server/tree/refactor/typescript-conversion). At least one typing looks like it should be defined/is dependent on @actual-app/api
> 
> @PartyLich I managed to miss this bullet point and ended up duplicating this work. I've rebased my migration branch on top of yours which also applies a number of linting fixes and typing improvements. Are you open to a PR onto your fork so we can merge these together?

I'm open to any path that gets the job done. 

My only concern about rebasing on a PR is the possibility that this PR *doesn't* get merged. In that case, my other two branches (and the future PRs that would result from them) as well as yours would be somewhere between useless and requiring a rewrite. My suggestion, perhaps misguided, is that we wait for the maintainer to return from their vacation. If/when this PR gets a review after that point, we can follow whatever path is easiest to get the best set of changes merged and continue improving the typing from there.

And sorry about the duplicated work! If you have any suggestions on how I should have written the description to avoid this, I'd love the chance to improve.

jlongster:
This is amazing, thank you @PartyLich. I'll review this PR soon

(hope you saw that I was out for a while which is why I haven't responded :) )

Dockerfile
```diff
@@ -1,15 +1,24 @@
...
+RUN yarn install && yarn build
+
 FROM node:16-bullseye as base
 RUN apt-get update && apt-get install -y openssl
 WORKDIR /app
 ENV NODE_ENV=production
-ADD yarn.lock package.json ./
-RUN npm rebuild bcrypt --build-from-source
-RUN yarn install --production
+COPY --from=build /app/build ./
```
jlongster, line 12: Why are there all of these Dockerfile changes in here?

These are unrelated to TypeScript. Please keep this PR scoped down as it will be a lot easier to review and merge in

jlongster: Unless I'm missing something; can't you just add `&& yarn build` here?

jlongster: And make all these other changes in another PR

jlongster (changes requested):
This looks great, thanks! Main thing is to reduce scope can we simplify the Dockerfile changes? If something breaks it feels better to have those changes in a separate PR

Otherwise this looks good. I just did some research and `ts-node` seems to be the most popular way to run TS in node? I'm surprised there isn't any formal guidance on TypeScript's website. It would be nice to reduce the number of steps though (we could run a build and run command in one, but still things could get out of sync). I'm in favor of ts-node if you want to try it, but maybe in a PR after this one?

jlongster:
I went ahead and merged #33. There are some minor conflicts, sorry but it shouldn't be too bad

PartyLich:
> Otherwise this looks good. I just did some research and `ts-node` seems to be the most popular way to run TS in node? I'm surprised there isn't any formal guidance on TypeScript's website. It would be nice to reduce the number of steps though (we could run a build and run command in one, but still things could get out of sync). I'm in favor of ts-node if you want to try it, but maybe in a PR after this one?

Just so I'm clear before I start making changes: we're aiming for ts-node in an additional PR, not in this?

jlongster:
Does it impact the docker image? If we used `ts-node` would we use that in the docker image to run it?

PartyLich:
Unless something has slipped my mind at the moment, ts-node would mean changing our start script in package.json, but leaves docker untouched. 

Iirc (2 weeks is a bit of time, sorry), the important docker changes were made to introduce a TS transpile step. With ts-node we wont need that, it transpiles on the fly.

PartyLich:
> Unless something has slipped my mind at the moment, ts-node would mean changing our start script in package.json, but leaves docker untouched.

scratch that. the docker file is using `node app.js` directly, so it would have to change.

jlongster:
> Iirc (2 weeks is a bit of time, sorry), the important docker changes were made to introduce a TS transpile step. With ts-node we wont need that, it transpiles on the fly.

Yeah, that's what I was talking about: we don't need all the `Dockerfile` changes anymore if we use `ts-node` right? If so, I'd say go ahead and integrate it with this PR

However, thinking about it more, it would be nice to stripe the type out once when generating the docker image, so the image doesn't have to run TS  every time is starts up. So let's use `ts-node` just for dev, but still build for prod in the docker image

And yes, with all the changes going on, please keep it as simple as possible and avoid restructuring other things in the Dockerfile in the same PR. It makes it harder to review :)

jlongster:
This looks good; looks like this makes it so we can run tsc on the files, but we're not actually changing anything to TS yet? When we do that, we will need to do 2 things:

* Integrate `ts-node` to run files in dev mode
* Change `Dockerfile` to compile out JS files which are then run (per my last comment, we should do this in prod to make startup time for the image faster)

PartyLich:
- conflicts with #33 resolved
- docker changes dropped from this PR. may be revisited later

If the path forward matches or includes these, could you give me some idea what order you'd like them submitted?

- add `ts-node`, and a `dev` (or other naming) script to run the server via ts-node 
- pre-build for the container, so prod isn't running typescript
- conversion of files to TS. Do you want them all at once (which is how i initially did it), broken up by file, something else?

PartyLich:
> This looks good; looks like this makes it so we can run tsc on the files, but we're not actually changing anything to TS yet? 

Correct.
 
Let me know if/how you want the other parts split, and you'll have those PRs pretty shortly.

jlongster:
Looks great! I'd say you can do 1 and 2 in the next PR (add `ts-node` for dev and compile out in docker), and after that we can start the migration.

For the migration, ideally we'd do it in several PRs for review sanity. The first PR should be "leaf" files like all the ones in `util`. After that, migrate each `app-` file separately. This will help keep PRs focused and able to land them, otherwise might be a lot of churn in a single PR.

(These PRs might not be that complicated they are just changing the extension to `ts` and adding a few types and using `any` where applicable. Even so, it's nice to do it in 3-5 PRs)

jlongster (approved):
Thanks a lot!

PR #33, opened 2022-05-15
Add more appropriate linting setup

TomAFrench:
Currently the linting setup is designed around a creating a React app and so isn't particularly useful for us. I've switched us over to use a Typescript compatible linting setup.

This does require a Typescript install however I'm expecting that we're going to have this anyway from #28 so it's not extra overhead.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,5 @@
+**/node_modules/*
+**/log/*
+**/shared/*
+
+supervise
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,6 +6,6 @@ module.exports = {
     node: true
   },
   parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint'],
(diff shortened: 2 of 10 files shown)


jlongster:
Sorry to use this as a test, but I'm setting up notifications so I need to make a comment on a PR.

jlongster:
Sounds like we're waiting until https://github.com/actualbudget/actual-server/pull/28 land?

TomAFrench:
There no strict dependency on #28 so they could be merged in either order. Just we probably wouldn't want to install typescript in this PR if we weren't planning on migrating the server over to typescript at some point.

jlongster:
Yep, that's fine with me. Was mainly thinking it will conflict with the other PR, but it shouldn't be too bad. I'll go ahead and merge this one to move things along and hopefully the author of the other can rebase & cleanup.

jlongster:
There are probably some rules I want to tweak, but we can do that later

PR #37, opened 2022-05-20
Add linter checks to CI

TomAFrench:
I forgot to add this to #33 but we should check the linting in CI on any PRs.


jlongster:
Nice!

PR #39, opened 2022-05-20
Build: tsnode for dev precompile for prod

PartyLich:
Adds:
- `dev` script that utilizes `ts-node`
- build step in the docker files

Concerns/Questions:
- `clean` script is using `rm -rf`, which is platform specific. I think it fails on like, Windows? We could use the `rimraf` package to get portable behavior.

Current image sizes:
|REPOSITORY |       TAG     |      IMAGE ID  |     CREATED  |      SIZE
|---|---|---|---|---|
|actual-server-alpine               |                            latest      |       3d72e9f7bd7b |  2 weeks ago    |     103MB
|actual-server           |                                    latest        |     6eb799e19345 |  2 weeks ago   |       241MB

Sizes using master branch (ie prior to these changes)
|REPOSITORY |       TAG     |      IMAGE ID  |     CREATED  |      SIZE
|---|---|---|---|---|
|actual-server-alp                      |                     old       |         334e2c0fb05a  | 22 seconds ago   |    110MB
|actual-server        |                                       old      |          a32605b18fd2  | About a minute ago  | 243MB

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "main": "index.js",
   "scripts": {
     "start": "node app",
+    "dev": "ts-node app",
     "lint": "eslint .",
     "build": "tsc",
     "types": "tsc --noEmit --incremental",
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "eslint": "^8.15.0",
     "eslint-plugin-prettier": "^4.0.0",
     "prettier": "^2.6.2",
+    "ts-node": "^10.9.1",
     "typescript": "^4.6.4"
   }
 }
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "dev": "ts-node app",
     "lint": "eslint .",
     "build": "tsc",
+    "clean": "rm -rf build",
     "types": "tsc --noEmit --incremental",
     "verify": "yarn -s lint && yarn types"
   },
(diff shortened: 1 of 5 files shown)


jlongster:
@PartyLich So sorry for the delay (again). Things have finally settled a bit for me so I'm going through these PRs, some of which are a bit old.

Just checking before I merge this -- you still good with everything here?

package.json
```diff
@@ -5,9 +5,11 @@
   "description": "actual syncing server",
   "main": "index.js",
   "scripts": {
-    "start": "node app",
+    "start": "node build/app",
+    "dev": "ts-node app",
     "lint": "eslint .",
     "build": "tsc",
+    "clean": "rm -rf build",
```
PartyLich, line 12: @jlongster adding a platform specific `rm -rf` to a script is not ideal. nothing else seems concerning atm.

jlongster: I think this is OK. We already assume an environment like git bash on windows. thanks!

README.md
```diff
@@ -14,7 +14,7 @@ It's very easy to get started. Clone this repo, install deps, and start it:
 git clone https://github.com/actualbudget/actual-server.git
 cd actual-server
 yarn install
-yarn start
+yarn dev
```
jlongster, line 17: Personally I think this is going to be too confusing to change, or at least not worth the confusion.

I'd prefer a `start:prod` or something that runs just the built sources, and leave `start` alone

PartyLich: adjusted.

PR #45, opened 2022-05-21
Refactor/ts conversion util dir

PartyLich:
#### Blockers
Based on #39. Will be rebased and reviewable if/when that pr makes progress. Until then it looks far larger than it is, because those commits are also present.
May also need to be updated after the prettier pr #43.

#### Details to note
- `handle-error.js` is only imported by `app-plaid.js`, which is currently unused. As such it hasn't been touched.
- `app-account.js` and `app-sync.js` changes will be gone as soon as those files are converted to ts. eg
`let errorMiddleware = require('./util/error-middleware').default;`
becomes
`import errorMiddleware from './util/error-middleware';`

- The lint rule changes are currently there to reflect the existing code (ie to minimize changes in the pr). Personally I'd prefer things like `prefer-const` enabled and fixed. If we can agree on that, we can make those changes (a bunch of `let` swapped for `const`) here or in a follow up.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
 node_modules
 user-files
 server-files
+tsconfig.tsbuildinfo
+Dockerfile*
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,9 @@ module.exports = {
   plugins: ['@typescript-eslint', 'prettier'],
   extends: ['eslint:recommended', "plugin:@typescript-eslint/recommended"],
   rules: {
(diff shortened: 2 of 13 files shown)


PartyLich:
There's a complete (probably?) [typescript branch](https://github.com/PartyLich/actual-server/tree/typescript) in my repo, and you're welcome to grab whatever you'd like from it. I'm not going to devote the time to splitting it up again. 

Leaving this PR open feels to me like it would be sending the wrong message.

PR #46, opened 2022-05-21
fix: error handling middleware signature

### LINKED ISSUE #40: Using localhost:5006 is not working
Screen Shot 2022-05-20 at 15 42 55
Screen Shot 2022-05-20 at 15 43 00

JazzaG:
Error handling middleware functions must have four arguments to
identify it as such.

Fixes #40

See https://expressjs.com/en/api.html

>Error-handling middleware always takes four arguments. You must provide four arguments to identify it as an error-handling middleware function. Even if you don’t need to use the next object, you must specify it to maintain the signature. Otherwise, the next object will be interpreted as regular middleware and will fail to handle errors.

### DIFF
--- a/util/error-middleware.js
+++ b/util/error-middleware.js
@@ -1,4 +1,4 @@
-async function middleware(err, req, res) {
+async function middleware(err, req, res, next) {
   console.log('ERROR', err);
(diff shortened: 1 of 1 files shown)


jlongster:
Thanks!

PR #50, opened 2022-05-28
Fix Cross-Origin issues to enable `SharedArrayBuffer`

jlongster:
This is the fix which enables an entirely different storage layer for Actual, and fixes all syncing/file issues with it. 

This took a while because I needed to get https://github.com/actualbudget/actual/pull/71 in first. I haven't released that change yet, but will do that this weekend. Once it's merged, I'll merge this PR.

You should export your data first before updating to this PR. The new storage layer will likely not be able to read your old data so it's really import that you'll be able to import your old data.

### DIFF
--- a/app.js
+++ b/app.js
@@ -29,7 +29,16 @@ app.get('/mode', (req, res) => {
 app.use(actuator()); // Provides /health, /metrics, /info
 
 // The web frontend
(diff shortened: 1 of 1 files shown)


jlongster:
Merging this since https://github.com/actualbudget/actual-server/pull/51 is in

PR #52, opened 2022-05-29
Fix 'Out of sync' error

7brend7:
When I've tried to modify some account (add a transaction) on **second** computer, I've got "Your data is out of sync" error on first.

image

This PR should fix sync error message.

NOT Done:
- handle the encryptKeyId  for encrypted messages;

### DIFF
--- a/app-sync.js
+++ b/app-sync.js
@@ -139,9 +139,16 @@ app.post('/sync', async (req, res) => {
   for (let i = 0; i < newMessages.length; i++) {
     let msg = newMessages[i];
     let envelopePb = new SyncPb.MessageEnvelope();
+    let messagePb = new SyncPb.Message();
+
+    messagePb.setDataset(msg.dataset);
+    messagePb.setRow(msg.row);
+    messagePb.setColumn(msg.column);
+    messagePb.setValue(msg.value);
+
     envelopePb.setTimestamp(msg.timestamp);
-    envelopePb.setIsencrypted(msg.is_encrypted === 1);
-    envelopePb.setContent(msg.content);
+    envelopePb.setContent(messagePb.serializeBinary());
+
     responsePb.addMessages(envelopePb);
   }
 
--- a/app-sync.js
+++ b/app-sync.js
@@ -136,21 +136,7 @@ app.post('/sync', async (req, res) => {
   let responsePb = new SyncPb.SyncResponse();
   responsePb.setMerkle(JSON.stringify(trie));
 
-  for (let i = 0; i < newMessages.length; i++) {
-    let msg = newMessages[i];
-    let envelopePb = new SyncPb.MessageEnvelope();
-    let messagePb = new SyncPb.Message();
-
-    messagePb.setDataset(msg.dataset);
-    messagePb.setRow(msg.row);
-    messagePb.setColumn(msg.column);
-    messagePb.setValue(msg.value);
-
-    envelopePb.setTimestamp(msg.timestamp);
-    envelopePb.setContent(messagePb.serializeBinary());
-
-    responsePb.addMessages(envelopePb);
-  }
+  newMessages.forEach(msg => responsePb.addMessages(msg));
 
   res.set('Content-Type', 'application/actual-sync');
   res.send(Buffer.from(responsePb.serializeBinary()));
--- a/sync-full.js
+++ b/sync-full.js
@@ -15,7 +15,7 @@ const sync = sequential(async function syncAPI(messages, since, fileId) {
     await actual.internal.send('load-budget', { id: fileId });
   }
 
-  messages = messages.map(envPb => {
+  const messagesDeserialized = messages.map(envPb => {
     let timestamp = envPb.getTimestamp();
     let msg = SyncPb.Message.deserializeBinary(envPb.getContent());
     return {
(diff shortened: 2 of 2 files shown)


bjornbostrom:
I have tried the 7brend7:fix-out-of-sync branch with some scenarios regarding accessing my budget files on several computers that did not work with the master branch. It seems to work perfectly OK 👍

bosmak:
It did fix the out-of-sync issue, but it creates some weird behaviors when using two or more files as they all become a merged copy of themselves.


[](https://gifyu.com/image/SxZJg)

bjornbostrom:
@bosmak: strange, when I tested yesterday I was actually testing it with two budgetfiles, and it all seems to work very smooth. When you had the weired behaivours, was it before @jlongster pushed the [Fix CORS issues](https://github.com/actualbudget/actual-server/commit/a2a460a8834517c50657a01e8f37997a3b2188a3) 2 days ago (hash a2a460a8834517c50657a01e8f37997a3b2188a3) ? The [Fix CORS issues](https://github.com/actualbudget/actual-server/commit/a2a460a8834517c50657a01e8f37997a3b2188a3) together with your PR might solve the problem altogether...?

bjornbostrom:
Strange... I also totally wiped the contents of the data-dir and uploaded  new fresh exported .zip's from the closed-source site. Did you also clean the data-dir? (just trying to figure out why it seems to work here...)

bosmak:
Yeah, I'm running on docker, So I built the new image and changed the mount points to a new folder to start a clean install.

bosmak:
@7brend7 @bjornbostrom I did some more testing, so what I found is, If I access from the localhost, it seems to work correctly, but when I use the IP, it does this strange behavior; in my case, I'm accessing the server on a different machine from the one hosting it.

So using `localhost` or `127.0.0.1` works without issue, but it happens as soon as I use `192.168.1.188`.

https://gifyu.com/image/SxrtG

bjornbostrom:
@bosmak I see, my setup is a virtual Rocky Linux 8.6 running docker and NGINX acting as a reverse proxy. So in my case, I am actually not browsing the docker-instance directly on port 5006. Rather I am browsing to https://my.server.com. The config in NGINX is 
`proxy_pass http://127.0.0.1:5006/;`

So I guess behind the scenes, NGINX is connecting to actual on localhost/127.0.0.1. 

Can that be the reason it works in my case then?

bosmak:
@bjornbostrom Interesting, I tried to replicate your NGINX reverse proxy setup, but no luck there.

But If I use a Cloudflare tunnel to access it externally it works correctly, with no issues and I can see they aren't duplicating anymore.

P.S. once a file was been "duplicated" it gets buggy as well, the name of the file is kept the same across all files, so I had to delete everything and start from scratch.

bjornbostrom:
Yep, I'm using https on my nginx.

jlongster:
@bosmak Whatever you have sitting in front of your server, it is dropping headers. You need to be sure that it is responding with these headers to enable `SharedArrayBuffer`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements

I'm going to be working on removing the "fallback" mode of browser storage because it's buggy. When this happens, the app will simply not work if `SharedArrayBuffer` is enabled so it will be more clear when this happens

app-sync.js
```diff
@@ -139,9 +139,16 @@ app.post('/sync', async (req, res) => {
...
+
+    messagePb.setDataset(msg.dataset);
+    messagePb.setRow(msg.row);
+    messagePb.setColumn(msg.column);
+    messagePb.setValue(msg.value);
+
     envelopePb.setTimestamp(msg.timestamp);
-    envelopePb.setIsencrypted(msg.is_encrypted === 1);
-    envelopePb.setContent(msg.content);
+    envelopePb.setContent(messagePb.serializeBinary());
```
jlongster, line 150: To support the `sync-simple.js` use case, where that flow that stores the binary-encoded messages directly, let's move encoding the messages into `sync-full.js`. `newMessages` should be an array of message envelopes, so the loop just becomes:

```
  for (let msg of newMessages) {
    responsePb.addMessages(msg);
  }
```

We'll also need to update the `sync-simple.js` flow to encode the messages, but I can do that

jlongster (changes requested):
Oh man, can't believe I didn't catch this 🤦 

Thanks so much for figuring this out! We definitely need to fix this to fix syncing across multiple devices. Dang, syncing should be robust but I missed some of this stuff when open-sourcing. At the core syncing is robust, but we just gotta make the server send back the right messages.

Let's move around the code a little bit. Thanks again!

jlongster:
@7brend7 thanks for the changes; do you see the option in the right sidebar to "allow edits from maintainers"? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

If so is it checked? I'm trying to figure out why I can push changes to PRs

7brend7:
@jlongster yep... it's checked... feel free to push any changes )

sdrichter:
FYI - You can recover from this issue by exporting your budget in Actual, then reimporting it on the patched version of the server. This restores all your accounts, transactions, etc. and the syncing issue does not come along with the imported data.

sdrichter:
FYI - No idea how I did it, but even running this PR on my server, my budget has somehow entered the syncing error state again and I can't get it to recover. So this patch doesn't seem to be foolproof.

bosmak:
@sdrichter do you have HTTPS configured? I always have the issue if I'm not using HTTPS

sdrichter:
I am using actual-server via Docker-Compose and running behind a Traefik reverse proxy. The actual-server instance itself is only running HTTP but I access it via HTTPS through the reverse proxy.

kinwolfqc:
I applied the PR last week, but today I got the sync error again, so, like @sdrichter , I can report that the problem is still there with the PR, it happens less often though.

jlongster:
Sorry about this! Finally going through PRs and this is the first one I'm merging because it fixes a critical problem.

I made a slight change to keep the changes small, but after talking with people we probably should start using `const` and not mutating variables to make it easier to use TypeScript.  I like using `let` everywhere, but that will be harder with TypeScript.

PR #63, opened 2022-07-30
build: add mount section to fly template

PartyLich:
Adds a mount section to the fly template for data persistence. It seems unlikely that there is a case in which one would want to deploy a syncing server but not provide that server with persistence. Providing a mount in the template seems like the saner default.

### DIFF
--- a/fly.template.toml
+++ b/fly.template.toml
@@ -11,6 +11,10 @@ processes = []
   allowed_public_ports = []
   auto_rollback = true
 
+[mounts]
+  source="actual_data"
+  destination="/data"
+
 [[services]]
   http_checks = []
   internal_port = 5006


jlongster:
@PartyLich I'm assuming that fly will error if you haven't created that volume yet, right?

You might be right that it's better to have this in there by default. I wanted to get people up and running as smoothly as possible, and then in the docs tell them how to persist. I'm fine with this though, but maybe we should simplify the docs some to more clearly state this. I can work on that

PartyLich:
> @PartyLich I'm assuming that fly will error if you haven't created that volume yet, right?

As far as I recall, if you try to deploy without creating the volume you'll receive an error. 

If that's troublesome, one can run the `fly launch` and choose not to deploy (which still creates the app/project, but doesnt start it). Then create the volume. Then run `fly deploy`. Which is the way I wrote it in [my fly+container guide](https://github.com/PartyLich/actual/wiki/Fly.io-image).

On the topic of docs, the wiki format did strike me as useful and easy for collaborators.

Issue #54, opened 2022-06-11
Crash - Running out of memory on fly.io when downloading the budget


SylvainMartel:
I am trying to download my budget on a new computer but the app keeps crashing on fly.io.  The only way I managed to make it work was to "sacle" the vm to 512mb for the download, then I was able to downscale it to 256 again.  Any ideas why the download seems to require more memory?(which fall out of the free tier if I'm correct.)

The app will crash-loop forever with 256Mb, here is the log from fly.io

```
 2022-06-11T13:17:24.258 runner[7c4ae106] yyz [info] Starting virtual machine
2022-06-11T13:17:24.453 app[7c4ae106] yyz [info] Starting init (commit: e3eb6d2)...
2022-06-11T13:17:24.467 app[7c4ae106] yyz [info] Mounting /dev/vdc at /data w/ uid: 0, gid: 0 and chmod 0755
2022-06-11T13:17:24.478 app[7c4ae106] yyz [info] Preparing to run: `/usr/bin/tini -g -- node app.js` as root
2022-06-11T13:17:24.504 app[7c4ae106] yyz [info] 2022/06/11 13:17:24 listening on [fdaa:0:5f82:a7b:8a0f:0:debd:2]:22 (DNS: [fdaa::3]:53)
2022-06-11T13:17:24.506 app[7c4ae106] yyz [info] [WARN tini (523)] Tini is not running as PID 1 and isn't registered as a child subreaper.
2022-06-11T13:17:24.506 app[7c4ae106] yyz [info] Zombie processes will not be re-parented to Tini, so zombie reaping won't work.
2022-06-11T13:17:24.506 app[7c4ae106] yyz [info] To fix the problem, use the -s option or set the environment variable TINI_SUBREAPER to register Tini as a child subreaper, or run Tini as PID 1.
2022-06-11T13:17:25.224 app[7c4ae106] yyz [info] Initializing Actual with user file dir: /data/user-files
2022-06-11T13:17:25.226 app[7c4ae106] yyz [info] Listening on 0.0.0.0:5006...
2022-06-11T13:19:28.352 app[7c4ae106] yyz [info] [ 123.971282] Out of memory: Killed process 528 (node) total-vm:947256kB, anon-rss:200868kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:1420kB oom_score_adj:0
2022-06-11T13:19:28.637 app[7c4ae106] yyz [info] Main child exited normally with code: 137
[...]

PR #64, opened 2022-07-30
build: add node GC argument to fly template

PartyLich:
Fly deployments on the free tier have ~256mb of memory available. Users with large transaction histories were encountering out of memory errors when attempting to export their data (#54).

This commit adds a node argument to (more or less) run the garbage collector at a smaller memory usage, helping keep users on flyio within their available limit. I think there are additional improvements to be made with the download process, but this config change is also useful.

### DIFF
--- a/fly.template.toml
+++ b/fly.template.toml
@@ -10,6 +10,7 @@ processes = []
 [experimental]
   allowed_public_ports = []
   auto_rollback = true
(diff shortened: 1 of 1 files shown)


jlongster (approved):

jlongster:
Nice find!!

PR #66, opened 2022-07-31
ci: add ARMv7 to image build

PartyLich:
A user reported a desire to run the application on hardware with an ARMv7 architecture. 

This change adds an additional container image to support that need.


sblcook:
This would be nice to run on a raspberry pi. Any way to bring this back?

PR #70, opened 2022-08-09
Update Express to version 4.17

### LINKED ISSUE #69: Sync Issue - Response has unsupported MIME type 'application/octet-stream' expected 'application/wasm'
An ongoing issue with syncronising budgets between devices still seems to be un-resolved, when loading budget a fallback message `falling back to ArrayBuffer instantiation` is presented. 

This initial message comes after `wasm streaming compile failed: TypeError: WebAssembly: Response has unsupported MIME type 'application/octet-stream' expected 'application/wasm'`

image

image

This appears to be due to an issue in Expresss - https://github.com/expressjs/express/issues/3589

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements

A bump in Express version appears to resolve the issue with fallback. 

Issues around sync can be found here - https://discord.com/channels/937901803608096828/937903364384096257/1006046258173059163 & here https://discord.com/channels/937901803608096828/937903364384096257/1006183778164691036 & here https://discord.com/channels/937901803608096828/937903364384096257/1006183934482190346

RichInSQL:
Updated express version to 4.17, this resolves #69

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,20 +19,20 @@
     "better-sqlite3": "^7.5.0",
     "body-parser": "^1.18.3",
     "cors": "^2.8.5",
-    "express": "^4.16.3",
+    "express": "4.17",
     "express-actuator": "^1.8.1",
     "express-response-size": "^0.0.3",
     "node-fetch": "^2.2.0",
     "uuid": "^3.3.2"
   },
   "devDependencies": {
+    "@types/better-sqlite3": "^7.5.0",
+    "@types/node": "^17.0.31",
     "@typescript-eslint/eslint-plugin": "^5.23.0",
     "@typescript-eslint/parser": "^5.23.0",
     "eslint": "^8.15.0",
     "eslint-plugin-prettier": "^4.0.0",
     "prettier": "^2.6.2",
-    "@types/better-sqlite3": "^7.5.0",
-    "@types/node": "^17.0.31",
     "typescript": "^4.6.4"
   }
 }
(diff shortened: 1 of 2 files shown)


trevdor:
Thought I'd provide some context to answer some questions @jlongster might have.

### Root Cause
`SharedArrayBuffer` wasm code fails its streaming compile because Express is serving `.wasm` as MIME type `application/octet-stream` instead of `application/wasm`. We fall back to the `ArrayBuffer`-based storage mechanism, which is known buggy.

Error screenshot (H/T @RichInSQL):
streaming_failed


### Why this Change Fixes Things
_tl;dr Not sure_

For reference: [express changelog](https://github.com/expressjs/express/releases)

I would have guessed the change that fixes our bug is the [removal of accepts@1.3.5](https://github.com/actualbudget/actual-server/pull/70/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL186). 
Leaves us depending on `accepts@1.3.8`, which depends on `mime-types@2.1.34` instead of `mime-types@2.1.18`.

But turns out `mime-types@2.1.18` is [the version where `application/wasm` was added](https://github.com/jshttp/mime-types/releases/tag/2.1.18) for `.wasm` files. So we should have already had the right content-type where we are seeing our bug. 

In any case, it feels safe to say the fix comes from bumping _some_ package in express' MIME types eco-system that we receive by going to `4.17.3` from `4.16.3`.

_Still think we ought to merge it. I just like to know exactly why whenever a version bump is the solution._  🤔

jlongster:
Looks great; looks like a safe bump. I'd be more worried about a major bump but this looks good. Thanks for the research!

PR #72, opened 2022-08-10
Fix/download only necessary files

PartyLich:
While investigating https://github.com/actualbudget/actual-server/issues/54 it was noted that the previous implementation zips the entire budget folder in the download endpoint. Once received on the client side, only the most recent db and metadata are actually used. This means up to 10 backups (~130mb in some cases) are being zipped in memory and transferred to the client (in addition to the two necessary files) despite none of that data being used. While this inefficiency isn't a major concern in some environments, it may be problematic in memory constrained environments (eg small flyio instances have 256mb total, and can consume half of that *at idle*).

This change transfers only the files that are actually utilized.

If there are other requests to this endpoint that *do* use the backups, this PR will have to be amended.

### DIFF
--- a/app-sync.js
+++ b/app-sync.js
@@ -330,7 +330,8 @@ app.get('/download-user-file', async (req, res) => {
 
   let zip = new AdmZip();
   try {
-    zip.addLocalFolder(join(config.userFiles, fileId), '/');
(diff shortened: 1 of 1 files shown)


jlongster (approved):
This is great, thank you!

I forgot that backups were a part of this. They definitely shouldn't be a part of it.

PR #81, opened 2022-10-06
Store user files as blobs instead of unzipping them

jlongster:
This is the last piece needed before releasing the new sync strategy. We should just write down the zip blobs directly instead of unzipping them; this also means we can turn back on e2e encryption if we want.

### DIFF
--- a/app-sync.js
+++ b/app-sync.js
@@ -1,12 +1,11 @@
+let fs = require('fs/promises');
 let { Buffer } = require('buffer');
-let { join } = require('path');
 let express = require('express');
 let uuid = require('uuid');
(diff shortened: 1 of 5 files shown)


jlongster:
Going to go ahead and merge this. Want to get somebody to test the image from master and then we'll do a release

SamRodkey:
Is there some kind of migration assistant or something where I can open files from 1.0.3 in the most recent releases? I keep getting "download errors" when I try to open my file. It's able to see it but not open it...

SamRodkey:
Found it here: https://actualbudget.github.io/docs/Getting-Started/migration/simple-sync

Sorry for the spam.

PR #29, opened 2022-05-05
Build and publish images with Github Actions

solidassassin:

PR #82, opened 2022-10-07
Build docker image on push to master or tag

trevdor:
~~Facilitates folks trying out any new merged changes, not just ones the project considers "ready for release."
Didn't include here, but we might also consider building on every git tag, since those are rare.~~

PR has changed a lot based on feedback. 
[See below for details](https://github.com/actualbudget/actual-server/pull/82#issuecomment-1276136334).

## Summary
Merging this change will result in Docker images with the following tags being built on every push to `master`:
`edge`
`sha-<commit_sha>`
(as well as the same tags suffixed with `-alpine`)

When a git tag containing `v*.*.*` is pushed to the repo, e.g. `v3.1.4` (or even `v1.2.3-long-description`, see Usage below), the following tagged Docker images will be built:
`latest`
`3.1`
`3.1.4`
`sha-<commit_sha>`
(as well as the same tags suffixed with `-alpine`)

## Usage
* Folks who want bleeding edge code can always use `actual-server:edge` 
* Folks who want a stable Actual can rely on `actual-server:latest`
* Anyone can lock to a specific commit at `actual-server:sha-<commit_sha>`
* Maintainers can push other tags, e.g. `v1.5.6-beta`, `v2.3.0-responsive-preview` as needed and images will be made available for testers

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,10 +1,32 @@
 name: Build Docker Image
 
-# Docker Images are only built when a new tag is created
+# Docker Images are built for every push to master or when a new tag is created
 on:
   push:
+    branches:
+      - master
     tags:
       - 'v*.*.*'
+    paths-ignore:
+      - README.md
+      - LICENSE.txt
+
+env:
+  IMAGES: |
+    jlongster/actual-server
+    ghcr.io/actualbudget/actual-server
+
+  # Creates the following tags:
+  # - actual-server:latest (see docker/metadata-action flavor inputs, below)
+  # - actual-server:edge (for master)
+  # - actual-server:1.3
+  # - actual-server:1.3.7
+  # - actual-server:sha-90dd603
+  TAGS: |
+    type=edge
+    type=semver,pattern={{version}}
+    type=semver,pattern={{major}}.{{minor}}
+    type=sha
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


denkristoffer:
This would make it easy to try out things like the new syncing, but I think most casual users would probably want a way to only use stable versions, so maybe we should have a Docker tag to differentiate between the two?

winklevos:
This is ideal,

I would say tags in `v*.*.*` format go to docker `:latest`

And `master` goes to `:develop`

trevdor:
Thanks for the suggestions. I welcome more input as I'm rather a GH Actions n00b. Might have even screwed up the syntax. I'll test that out on my own fork later today.

Here are the improvements I made:

1. Consolidate reused values as `env` vars 
2. Separate stable and unstable lineages with these tags:
  * `edge` - built from `master`
    From [the tagging workflow we're using](https://github.com/marketplace/actions/docker-metadata-action#typeedge):
    > An edge tag reflects the last commit of the active branch on your Git repository.
  * semver (e.g. `1.2.3`, `1.2`) - built from `v*.*.*` tags
  * `latest` - auto-updated to point to the most recent semver build (see [the workflow docs](https://github.com/marketplace/actions/docker-metadata-action#typeedge))
  * sha (e.g. `sha-ad132f5`) - built for each build above
3. Ignored a couple non-code files, borrowing from https://github.com/actualbudget/actual-server/pull/29
    
Still only building on push to `v*.*.*` tags and `master`.

trevdor:
@jlongster I'm feeling good about this workflow. Mind merging or giving me commit access to get this in? Most immediately, it'll get us the missing `master` image so folks can easily try out the sync and blob storage fixes.

ljford7:
I apologize if this is the wrong place, but the example docker compose file here: [https://raw.githubusercontent.com/actualbudget/actual-server/master/docker-compose.yml](https://raw.githubusercontent.com/actualbudget/actual-server/master/docker-compose.yml) is missing the "image: jlongster/actual-server:latest". Once I added it, it works great.

trevdor:
> I apologize if this is the wrong place, but the example docker compose file here: https://raw.githubusercontent.com/actualbudget/actual-server/master/docker-compose.yml is missing the "image: jlongster/actual-server:latest". Once I added it, it works great.

@ljford7 Can you open a PR with your change where we could discuss it further?

PR #85, opened 2022-10-25
Version bump 4.1.1

rich-howell:
Actual Web version bump to pull latest version from NPM


jlongster:
Thanks for this! I was going to update this PR but because I've changed the version I opened a new PR since the branch name is based on the version: https://github.com/actualbudget/actual-server/pull/86

PR #89, opened 2022-10-29
Allow config.json to contain a partial config

nfaucher8:
This is a small PR that changes how `load-config.js` loads a config from `config.json`. This would allow for `config.json` to contain "partial" configs.

Example `config.json`
```
{
    "port": 5007,
    "hostname": "127.0.0.1"
}
```
This would override the `config.port` and `config.hostname` but `config.userFiles` and `config.serverFiles` would be defaulted.


MatissJanis (approved):
👋 Sorry for the slow review, but this looks good! LGTM!

PR #95, opened 2022-12-08
Don't public major.minor format of docker images

jlongster:
We don't want `22.12`, we always want the full versioon


jlongster:
Duplicate of https://github.com/actualbudget/actual-server/pull/94

PR #102, opened 2023-01-11
fix: :lock: Password change now invalidates old token

### LINKED ISSUE #101: Password Change should Invalidate token
In the /change-password endpoint, once the password is changed, it is a good practice to drop the existing token and generate a new token.

In the current scenario, the updated password will only be needed for new clients but the existing clients will continue to be able communicate.

Implementing this will force all the other clients if any to re-login with a new password.

urjeetpatel:
Password change will create a new token in the database. This will fix #101


app-account.js
```diff
@@ -94,12 +94,13 @@ app.post('/change-password', (req, res) => {
...
   let hashed = hashPassword(password);
-
+  let token = uuid.v4();
   // Note that this doesn't have a WHERE. This table only ever has 1
   // row (maybe that will change in the future? if this this will not work)
   accountDb.mutate('UPDATE auth SET password = ?', [hashed]);
+  accountDb.mutate('UPDATE sessions SET token = ?', [token]);
 
-  res.send({ status: 'ok', data: {} });
+  res.send({ status: 'ok', data: { token } });
```
MatissJanis, line 103: **❓ question:** ‏why are you returning the `token` here? Is it actually used in the UI?

MatissJanis (commented):
👋 Sorry for the super slow reply. My goal was to first get the [Actual](https://github.com/actualbudget/actual) PRs under control and then look at actual-server.

Overall IMO this looks good. But I'm interested to hear the response to my question before merging.

urjeetpatel:
I do not believe that the UI is expecting a new token just yet. but if the token is invalidated on the backend, it is a good practice to send the ui the new token or else the UI will be kicked out on the next request as unauthenticated.

MatissJanis:
> I do not believe that the UI is expecting a new token just yet. but if the token is invalidated on the backend, it is a good practice to send the ui the new token or else the UI will be kicked out on the next request as unauthenticated.

Sure, but will it be used in the UI? If you plan on submitting a PR there too, then I'm happy to review it too. But if you're not planing to do the UI change, then I'd err on the side of caution and simplicity. Thus **not** to return it from the API unless it's actually being used.

urjeetpatel:
I took a quick peek at the UI code and it is not being used. also I am not a veteran react developer so it is not in my capabilities to help enhance the ui to use it. i will push a new commit to remove the token in the return.

urjeetpatel:
ok please review now.

MatissJanis (approved):
Amazing! Just tested it locally and it works as expected.

- active user gets logged out after changing password
- other users also get logged out (but they are not redirected to login page; they can continue using the app offline with a small banner notifying them of the lost connectivity).

Thanks for your contribution!

PR #109, opened 2023-01-30
fix: listen also listen on ipv6 addresses

### LINKED ISSUE #108: Not listening on all addresses by default
The hostname in the default configuration is `0.0.0.0` which will only listen on all available IPv4 addresses. This means the default configuration will not work in IPv6 only environments and work only unreliably in dual stack environments (such as many hosting providers).

The hostname in the default configuration should be changed to `::` which will listen on both IP stacks. The notation might suggest that this would make it IPv6 only but this is not the case.

heilerich:
Fixes #108 

Listening on '::' will make node listen on IPv4 as well as IPv6 addresses.

### DIFF
--- a/load-config.js
+++ b/load-config.js
@@ -14,7 +14,7 @@ if (process.env.NODE_ENV === 'test') {
   config = {
     mode: 'test',
     port: 5006,
-    hostname: '0.0.0.0',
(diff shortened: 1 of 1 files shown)


MatissJanis (approved):
Thanks!

PR #114, opened 2023-02-01
[WIP] :building_construction:  support typescript code

MatissJanis:
A few changes in this PR to facilitate migration to TypeScript.

1. Tooling setup to allow developing with TS files
2. Changing imports/exports to be more modern
3. Changing some file paths now that files are served from `./build` folder instead of `.`

**Why do this?**

Upgrading some dependencies with security issues (such as node-fetch) requires using modern import syntax. So we might as well continue the TS migration here..

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,12 +1,13 @@
-const fs = require('fs');
-const express = require('express');
-const actuator = require('express-actuator');
-const bodyParser = require('body-parser');
-const cors = require('cors');
-const config = require('./load-config');
+import fs from 'fs';
+import path from 'path';
+import express from 'express';
+import actuator from 'express-actuator';
+import bodyParser from 'body-parser';
+import cors from 'cors';
+import config from './load-config';
 
-const accountApp = require('./app-account');
-const syncApp = require('./app-sync');
+import * as accountApp from './app-account';
+import * as syncApp from './app-sync';
 
 const app = express();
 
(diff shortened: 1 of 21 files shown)


MatissJanis:
Note: I did not yet test actually deploying the build to pikapods/fly.io or anywhere else. Need to check if this still works as expected before merging.

MatissJanis:
Also: maybe this is an overkill. Still need to consider alternatives.

But either way we need to figure out the import issue: https://github.com/node-fetch/node-fetch/blob/HEAD/docs/v3-UPGRADE-GUIDE.md#converted-to-es-module

app.js
```diff
@@ -35,12 +36,19 @@ app.use((req, res, next) => {
...
 });
 app.use(
-  express.static(__dirname + '/node_modules/@actual-app/web/build', {
-    index: false
-  })
+  express.static(
+    path.resolve(__dirname + '/../node_modules/@actual-app/web/build'),
+    {
+      index: false
+    }
```
j-f1, line 43:
```suggestion
    { index: false }
```

app.js
```diff
@@ -35,12 +36,19 @@ app.use((req, res, next) => {
...
+  )
 );
 app.get('/*', (req, res) => {
-  res.sendFile(__dirname + '/node_modules/@actual-app/web/build/index.html');
+  res.sendFile(
+    path.resolve(
+      __dirname + '/../node_modules/@actual-app/web/build/index.html'
+    )
+  );
```
j-f1, line 51: Looking at the check failure (which is probably safe to ignore like you did), we could pull the file out into a top level `fs.readFileSync` and then just send back the buffer each time.

jest.setup.js
```diff
@@ -1,7 +1,7 @@
 const fs = require('fs');
 const { join } = require('path');
 const { getAccountDb } = require('./account-db');
-const config = require('./load-config');
+const { default: config } = require('./load-config');
```
j-f1, line 4: Is it possible to switch this file to ESM too? real ESM files can’t be `require()`-ed synchronously.

j-f1 (approved):
This would be cool! I have a couple of specific thoughts below. Also, IMO we should upgrade to Node 19 (I was just getting started on that but hadn’t actually done anything yet) so that we can use the `fetch()` global and other newer features/security improvements.

j-f1 (commented):
oops wrong button

PR #115, opened 2023-02-01
Add support for HTTPS

j-f1:
This description is out of date, the PR just adds HTTPS support.

<details>
I’ve also taken the opportunity to switch from Express to Fastify here, and enable HTTP/2 support when HTTPS is used.

Express seems to be slipping out of date, and Fastify has a much more rapid release cadence. I was first introduced to it via [Glitch](https://glitch.com), but I’m also happy to revert the commits relating to HTTP/2 and Fastify if you’d rather stick to the tried-and-true.

Either way, I’d recommend reviewing the first commit separately from the others since that has the HTTPS changes.


MatissJanis:
👋 I tried testing this locally, but it does not seem to be working as expected.

What I expect to work (however it does not work out of the box):
```
yarn install
yarn start
# open https://localhost:5006/ in chrome
```

--

But if we take that aside: this PR has a high risk factor. Not only we are introducing a new technology and replacing the previous one, but also we are shifting a lot of code around and changing things. This gives me low-confidence in the stability here.

What benefit does Fastify give over Express? If it's purely because one is "cooler" than the other - I'd prefer to stick with Express as we're confident in the stability of it.

j-f1:
> What I expect to work (however it does not work out of the box):

Did you configure a certificate and key? If not, it will start a normal HTTP server on port 5006 instead.

> But if we take that aside: this PR has a high risk factor. Not only we are introducing a new technology and replacing the previous one, but also we are shifting a lot of code around and changing things. This gives me low-confidence in the stability here.
> 
> What benefit does Fastify give over Express? If it's purely because one is "cooler" than the other - I'd prefer to stick with Express as we're confident in the stability of it.

That’s fair! It seems like if we want to support newer web technologies like HTTP/2 and HTTP/3 (which will have support coming to Node eventually™), Fastify seems to be more invested in getting onto those updates while Express still hasn’t adopted support for HTTP/2 after several years. A couple of the maintainers of Fastify are on the Node.js leadership committee, whereas that doesn’t appear to be the case for Express. That said, I don’t think having support for the latest web technologies is that important for us in general, so I’d be happy to revert the migration and instead stick with HTTPS on HTTP/1.1.

I’ll also whip up a PR to the docs site that describes how to get HTTPS working.

MatissJanis:
> Did you configure a certificate and key? If not, it will start a normal HTTP server on port 5006 instead.

Nope. Just opening the same page as before on a fresh install.

MatissJanis (approved):

PR #116, opened 2023-02-02
:fire:  remove unused code (plaid, sync-full)

MatissJanis:
Removing dead (unused) code.

The less code we have to maintain - the better. Plus this allows to remove some dependencies.

We can always bring this back when someone re-implements bank sync integration. I've linked to the deleted file [here](https://github.com/actualbudget/actual/issues/454) so we won't lose it.


j-f1 (approved):
Great! 🔥

PR #118, opened 2023-02-02
:arrow_up:  upgrade @actual-app/api to 4.1.6: node-fetch v2 support

MatissJanis:
Upgrading `@actual-app/api` to 4.1.6: this will also upgrade node-fetch to v2 and thus fix security problems we have here.

Upgrade PR in api: https://github.com/actualbudget/actual/pull/609

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "verify": "yarn -s lint && yarn types"
   },
   "dependencies": {
-    "@actual-app/api": "4.1.5",
+    "@actual-app/api": "^4.1.6",
     "@actual-app/web": "23.1.12",
     "bcrypt": "^5.0.1",
     "better-sqlite3": "^7.5.0",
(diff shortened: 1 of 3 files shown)


MatissJanis:
Tested locally. Works as expected

package.json
```diff
@@ -13,7 +13,7 @@
     "verify": "yarn -s lint && yarn types"
   },
   "dependencies": {
-    "@actual-app/api": "4.1.5",
+    "@actual-app/api": "^4.1.6",
```
j-f1, line 16: Can you pin this dependency back? Want to make sure previous versions don’t accidentally break with future releases of the API

```suggestion
    "@actual-app/api": "4.1.6",
```

MatissJanis: Makes sense. I'll also slip in `.editorconfig` addition here :)
https://github.com/actualbudget/actual-server/pull/118/commits/f070ddca905bbca93985cfa86cc134e5793ceb0c

j-f1 (approved):
👍

PR #120, opened 2023-02-04
Improve the server config

j-f1:
- Now all config properties can be set via environment variables (will PR the docs to add them there too!)
- a new `webRoot` config property allows you to skip using the web UI installed from npm and instead use a different version. The only requirement is that it has an `index.html`.

### DIFF
--- a/load-config.js
+++ b/load-config.js
@@ -10,21 +10,22 @@ try {
   // do nothing
 }
 
+let defaultConfig = {
+  port: 5006,
+  hostname: '::',
+  serverFiles: join(root, 'server-files'),
+  userFiles: join(root, 'user-files'),
+  webRoot: join(__dirname, 'node_modules', '@actual-app', 'web', 'build')
+};
 if (process.env.NODE_ENV === 'test') {
   config = {
     mode: 'test',
-    port: 5006,
-    hostname: '::',
-    serverFiles: join(__dirname, 'test-server-files'),
-    userFiles: join(__dirname, 'test-user-files')
+    ...defaultConfig
   };
 } else {
   config = {
     mode: 'development',
-    port: 5006,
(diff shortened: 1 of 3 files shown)


j-f1:
Not sure why the tests are failing

load-config.js
```diff
@@ -1,35 +1,52 @@
...
+
+/** @type {import('./config-types').Config} */
+let config;
 if (process.env.NODE_ENV === 'test') {
   config = {
     mode: 'test',
-    port: 5006,
-    hostname: '::',
-    serverFiles: join(__dirname, 'test-server-files'),
-    userFiles: join(__dirname, 'test-user-files')
```
MatissJanis, line 19: Should leave these two in. IMO that's why the tests are failing.

load-config.js
```diff
@@ -1,35 +1,52 @@
...
   };
 }
 
-// The env variable always takes precedence
-config.userFiles = process.env.ACTUAL_USER_FILES || config.userFiles;
-
-module.exports = config;
+module.exports = {
+  ...config,
+  port: +process.env.ACTUAL_PORT || +process.env.PORT || config.port,
```
MatissJanis, line 39: **👏 praise:** ‏good idea to add these as env vars. Should we also use `dotenv` to pick automatically pick these up from `.env` file? And then we could also create a `.env.example` file where we list them all out and provide a basic documentation. WDYT?

j-f1: IMO it’s more ergonomic to use the `config.json` if you can, this is intended for Docker and such where it’s easy to set env vars but hard to inject a specific file. (For example in my other PR I use it to override the web root for the `edge` images).

MatissJanis (approved):

MatissJanis (approved):

PR #56, opened 2022-06-18
Migrate to yarn v3.2.1

TomAFrench:
This PR migrates this repo to use yarn v3 for consistency with https://github.com/actualbudget/actual/pull/50.

PR #121, opened 2023-02-04
Update the :edge images to use the very latest web UI version

j-f1:
- The `:edge` and `:edge-alpine` containers will now download the latest commit of the frontend repo and build it when they are built.
- The backend dependency install process now uses the local copy of Yarn (from #56) to install dependencies
- Extraneous files are no longer copied into the production image. This is mostly a win for local builds (where the `.yarn` cache folder used to get copied in) but it’s also a small win for CI-built containers which will no longer have the README/configuration files/Yarn source code lying around
- The dockerfiles now say `EXPOSE 5006`. This doesn’t have an effect on anything by default as far as I can tell but it does mean that tools know we will be listening on that port and can help users set up their container accordingly (for example, Docker Desktop)
- Remove some extraneous dependencies from the containers (including openssl)


.github/workflows/docker-edge.yml
```diff
@@ -0,0 +1,82 @@
...
+    name: Build Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
```
trevdor, line 35:
```suggestion
        uses: docker/setup-qemu-action@v2

      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v2
```

.github/workflows/docker-edge.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Docker meta for Alpine image
+        id: alpine-meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ env.IMAGES }}
+          flavor: suffix=-alpine
+          tags: $${{ env.TAGS }}
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
```
trevdor, line 54:
```suggestion
        uses: docker/login-action@v2
```

.github/workflows/docker-edge.yml
```diff
@@ -0,0 +1,82 @@
...
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          file: docker/edge-ubuntu.Dockerfile
+          platforms: linux/amd64,linux/arm64
+          tags: ${{ steps.meta.outputs.tags }}
+
+      - name: Build and push Alpine image
+        uses: docker/build-push-action@v2
```
trevdor, line 76:
```suggestion
        uses: docker/build-push-action@v4
        with:
          context: .
          push: true
          file: docker/edge-ubuntu.Dockerfile
          platforms: linux/amd64,linux/arm64
          tags: ${{ steps.meta.outputs.tags }}

      - name: Build and push Alpine image
        uses: docker/build-push-action@v4
```

trevdor (commented):
Super cool!
Not clicking "Approve" only bc I haven't tested the images output by your new Dockerfiles. 

Oh, and nitpicked all over your GH actions versions. 😛 
Just saw your node 18 update so I know you like new stuff. 😆

j-f1:
Updated!

.github/workflows/docker-release.yml
```diff
@@ -1,10 +1,8 @@
-name: Build Docker Image
+name: Build Stable Docker Image
 
-# Docker Images are built for every push to master or when a new tag is created
+# Stable Docker images are built for every new tag
```
trevdor, line 3: Your comment update makes me think: maybe we ought to truly build for every new tag? Instead of just `v*.*.*`. After all, why would anyone with permission push a tag we didn't want someone to be able to try out? 🤔

j-f1: IMO we should keep the `:latest` channel to official releases (e.g. `vX.Y.Z`) so people don’t unintentionally get bumped to an unstable version. It might make sense to set that up in the future but we can always figure that out when we need it :)

docker/edge-alpine.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM alpine as base
```
trevdor, line 1: I'm curious why `alpine:latest` instead of picking a version? 
Thinking about the potential for a build to break in the future if anything in our image depends on something that should happen to change as alpine pushes forward over time.

docker/edge-alpine.Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ADD .yarn ./.yarn
+ADD yarn.lock package.json .yarnrc.yml ./
+RUN yarn workspaces focus --all --production
+
+# Since we’re just using static files, use the Ubuntu image to build the frontend
+# (otherwise electron fails to build)
+FROM node:16-bullseye as frontend
+WORKDIR /frontend
+# Rebuild whenever there are new commits to the frontend
+ADD "https://api.github.com/repos/actualbudget/actual/commits" /tmp/actual-commit.json
```
trevdor, line 13: Neat cross-repo trick!

docker/edge-alpine.Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN git clone --depth=1 https://github.com/actualbudget/actual /frontend
```
trevdor, line 14: Hmm...could be docker magic I don't know...but is this URL correct? If so, how?? 😸

j-f1: What do you mean?

src/load-config.js
```diff
@@ -2,9 +2,8 @@ import fs from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 
-export const projectRoot = path.dirname(
-  path.dirname(fileURLToPath(import.meta.url))
-);
+const projectRoot = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
+export const sqlDir = path.join(projectRoot, 'src', 'sql');
```
trevdor, line 6: At first glance, this feels like a change in path.
For example, [up here](https://github.com/actualbudget/actual-server/pull/121/files#diff-95e90f57d7a98dcb3cee08f7082b6dfcbd59d1205bc80d5766cc5a8e159e38c5L21) isn't the old version `<projectRoot>/sql/account.sql` and the new `<projectRoot>/src/sql/account.sql`?

If I have that right, is that intentional?

j-f1: Yes, I moved the .sql files into `src/` so that the containers only have to copy `app.js` and the `src` folder into the container (plus `package.json` and `yarn.lock` for dependencies)

trevdor (approved):
🥳 This is a feature I've wanted from our CI for as long as I can remember.
And thanks for the housekeeping work of slimming down these images.

Left some questions & comments, but deal with 'em as you see fit and ship it.

One meta-comment: the changes around `sqlDir` in [src/load-config.js](https://github.com/actualbudget/actual-server/pull/121/files#diff-6cebf5e0cbfe01dc18e7db84c32b0f62dbdce59922112474c045dc9b2fc42cdb) and surrounding appear off-topic for this PR at first glance. I'd suggest dropping them to keep the work focused if they are, in fact, unrelated.

docker/edge-alpine.Dockerfile
trevdor: Oh shoot -- it wrapped just right in side-by-side mode so that I thought I was seeing `https://github.com/actualbudget/actual/frontend` rather than two arguments. 
Ignore me!

src/load-config.js
trevdor: Ah, I see how that's related now. That answers my approval comment meta-question. Thanks.

BoxedBrain:
@j-f1 appreciate your work here! Just want to mention, that this breaks the `Running via Docker` guide atm.

j-f1:
Oh no! Which part of the guide is not working for you?

PR #131, opened 2023-02-07
Move source code to an src/ directory

j-f1:
This cleans up the root folder and makes it easier to differentiate between code and support/scaffolding. It *should* be compatible with all existing things that do `node app.js` to start the app (although we should perhaps encourage people to do `npm start` instead.


MatissJanis (approved):
Thanks!

PR #133, opened 2023-02-11
Replace require with import

j-f1:
This fixes an issue where the server would fail to start up when HTTPS was enabled, since ES Modules no longer support calling `require()` inside of them directly. I checked the rest of the code and there don’t seem to be any other `require` calls.


trevdor (approved):
👍🏻

PR #135, opened 2023-02-11
Auto-configure the client’s server URL

j-f1:
Corresponding PR for https://github.com/actualbudget/actual/pull/649. Also allows specifying relative paths for `webRoot` (such as `../actual/packages/desktop-client/build`, which is useful if you want to test running against a local build of the frontend).

### DIFF
--- a/src/app-account.js
+++ b/src/app-account.js
@@ -30,10 +30,7 @@ app.get('/needs-bootstrap', (req, res) => {
 
   res.send({
     status: 'ok',
-    data: {
-      isActual: true,
-      bootstrapped: rows.length > 0
-    }
+    data: { bootstrapped: rows.length > 0 }
   });
 });
 
--- a/src/app-account.js
+++ b/src/app-account.js
@@ -20,6 +20,7 @@ function hashPassword(password) {
 
 // Non-authenticated endpoints:
 //
+// /needs-bootstrap
 // /boostrap (special endpoint for setting up the instance, cant call again)
 // /login
 
--- a/src/app-account.js
+++ b/src/app-account.js
@@ -29,7 +29,10 @@ app.get('/needs-bootstrap', (req, res) => {
 
   res.send({
     status: 'ok',
-    data: { bootstrapped: rows.length > 0 }
+    data: {
+      isActual: true,
+      bootstrapped: rows.length > 0
+    }
   });
 });
 
--- a/src/app.js
+++ b/src/app.js
@@ -22,6 +22,13 @@ app.use(bodyParser.raw({ type: 'application/encrypted-file', limit: '50mb' }));
 app.use('/sync', syncApp.handlers);
 app.use('/account', accountApp.handlers);
 
+app.get('/client-bootstrap', (req, res) => {
+  // can also return a `serverURL` to override the server URL
+  // (but our server doesn’t need that because it will default
+  // to the client’s domain which is fine in our case)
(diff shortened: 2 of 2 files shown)


MatissJanis:
Why not use the existing `/needs-bootstrap` for this? It's used to check if the server is running actual-server or a different server after typing in the URL and clicking on "ok".

src/load-config.js
```diff
@@ -54,7 +54,10 @@ export default {
   hostname: process.env.ACTUAL_HOSTNAME || config.hostname,
   serverFiles: process.env.ACTUAL_SERVER_FILES || config.serverFiles,
   userFiles: process.env.ACTUAL_USER_FILES || config.userFiles,
-  webRoot: process.env.ACTUAL_WEB_ROOT || config.webRoot,
+  webRoot: path.resolve(
+    projectRoot,
+    process.env.ACTUAL_WEB_ROOT || config.webRoot
+  ),
```
MatissJanis, line 60: **🔨 warning:** ‏the more I think about this, the more I dislike it from a security POV.

If an attacker gains access to this config file or the env variables: he basically can traverse the entire filesystem. I imagine setting it to `../../../path/to/budget.data?ignore=` and then opening up the server would spit out the budget file. 

https://www.acunetix.com/websitesecurity/directory-traversal/

MatissJanis:
> > Why not use the existing `/needs-bootstrap` for this? It's used to check if the server is running actual-server or a different server after typing in the URL and clicking on "ok".
> 
> That’s an interesting idea! So you’re thinking we would call `/needs-bootstrap` on the current domain immediately at startup?

yup

src/load-config.js
j-f1: If the attacker gains access to the config file then they presumably already have access to the filesystem, right?

And the environment variables are evaluated once at startup, so changes to them at runtime would not have any effect. If you were able to set them before server startup, you could already change the `ACTUAL_SERVER_FILES` environment variable to an empty directory, then re-bootstrap the app and download all of the files with the normal UI.

MatissJanis: Not necessarily. There might be another system running on the same instance. And the attacker might be able to upload files to it, but not read from it.

Anyways, that's a lot of _ifs_. However, we do need to consider them and try and keep the attack vector for actual-server as little as possible. Directory traversal attack is a pretty common security issue. So if we can treat **all** the input as potentially malicious - we can protect ourselves better.

In this case I really don't like the wrapping of this input in `path.resolve`. Is there an alternative we could consider that would solve your problem? Perhaps symlinks?

j-f1:
Since d39f196b8702de79c448752b80956de0c57adabe needs more discussion, I’ll move that to a different PR and focus this on the issue at hand.

src/app-account.js
```diff
@@ -29,7 +29,10 @@ app.get('/needs-bootstrap', (req, res) => {
 
   res.send({
     status: 'ok',
-    data: { bootstrapped: rows.length > 0 }
+    data: {
+      isActual: true,
```
MatissJanis, line 33: Do we really need this constant? Wouldn't a 200 response from this endpoint be sufficient to tell if it's an actual server?

j-f1: Sure.

MatissJanis (approved):

j-f1:
Will wait to merge until the other PR is also approved to make sure there don’t need to be any other changes

PR #138, opened 2023-02-14
Support webRoot config variable being a relative path

j-f1:
Note: See earlier discussion about this in https://github.com/actualbudget/actual-server/pull/135#discussion_r1106295533, need to think about the security implications of this before merging.


MatissJanis:
Is it just for local testing? Did you consider using `yarn link` instead?

j-f1:
Absolute paths are fine.

PR #141, opened 2023-02-17
Add linux/arm/v7 to the list of platforms for the Docker image

jamesmortensen:
Users in #99 report that Actual in Docker runs on armv7 platforms, although a bit sluggish.  I confirmed that the base images for Debian and Alpine support the linux/arm/v7 target and have added them to the platform list in the GitHub Actions workflow.  

At least one user confirms it works with the bullseye default `Dockerfile`, but before merging it would be great if someone can confirm it works with the Dockerfile.alpine image:

```
git clone https://github.com/jamesmortensen/actual-server.git
cd actual-server
git checkout armv7-image
docker build -t actual-server -f Dockerfile.alpine .
docker run -p 5006:5006 actual-server
```


SalvyTek:
I built the docker image on my Raspberry Pi 4 (armv8A) and everything seems to work. I don't have a device with armV7. I am sorry,

jamesmortensen:
@SalvyTek No worries.  It sounds like there are others who can validate for armv7.

j-f1 (approved):
Looks good to me, thanks for contributing!

PR #144, opened 2023-03-04
:sparkles: (bank-sync) meaningful default balance calculation

MatissJanis:
Set a default balance calculation logic. This will allow us to integrate with the majority of the Nordigen supported banks without needing to manually implement each individual bank.

Frontend change: https://github.com/actualbudget/actual/pull/727

**Important**: some banks might break.. sadly I cannot test all of the banks. But we should successfully cover ~80-90% of the supported banks with these defaults. And if people notice that some specific bank is broken - they can alwys report it to us and we can try and implement a custom bank class for them.


MatissJanis:
Currently tested this with the 2x banks that I have: N26 and Swedbank (Latvia). Please feel free to test this with more banks and let me know what you think. If all goes to plan, IMO we will be able to merge this.

MatissJanis:
Bank list: https://docs.google.com/spreadsheets/d/1ogpzydzotOltbssrc3IQ8rhBLlIZbQgm5QCiiNJrkyA/edit#gid=489769432

src/app-nordigen/banks/integration-bank.js
```diff
@@ -33,6 +33,10 @@ export default {
         top10SortedTransactions: JSON.stringify(sortedTransactions.slice(0, 10))
       }
     );
-    return 0;
+
+    const currentBalance = balances[0];
```
fstybel, line 37: Let's use here the priority of balance types. WDYT? IMO we should base if it's possible on the booked transactions, not on which are expected.
https://nordigen.com/en/docs/account-information/output/balance/

MatissJanis: Updated the PR.

Is this what you meant?

fstybel: Yes, I'm not sure if it's the correct order but I'm not an expert in the account types :D I believe we will have a possibility to tweak it if needed.
IMO the most important is to have at the end of this list balance types based on the not booked transactions because they can just disappear in the meantime. On another side, we use this function only during adding a new account.

fstybel (approved):

src/app-nordigen/banks/integration-bank.js
MatissJanis: I just double checked with my Latvian banks and it turns out we do need `expected` balanceType in the list for things to work. Without it my balance is calculated totally wrong.

https://github.com/actualbudget/actual-server/pull/144/commits/594df22a9d2a6f1d7e14fe41e7e26c744c6b2834

fstybel (approved):
Looks good!

j-f1 (approved):

PR #145, opened 2023-03-05
:sparkles: (bank-sync) expose available banks via an endpoint

MatissJanis:
Related to: https://github.com/actualbudget/actual/pull/727

Exposing the available banks via an endpoint. Currently only the "fully" supported banks are exposed, but very soon the plan is to expose all of the bank by default (as part of https://github.com/actualbudget/actual-server/pull/144)


src/app-nordigen/banks.js
```diff
@@ -0,0 +1,4753 @@
+// https://docs.google.com/spreadsheets/d/1ogpzydzotOltbssrc3IQ8rhBLlIZbQgm5QCiiNJrkyA/edit#gid=489769432
```
fstybel, line 1: Maybe a better way will be to fetch the available banks from the Nordigen endpoint. https://nordigen.com/en/docs/account-information/integration/parameters-and-responses/#/institutions/retrieve%20all%20supported%20Institutions%20in%20a%20given%20country

MatissJanis: oo.. I like this idea!

So I'll do this in 2 parts (in order to unblock the frontend PR: https://github.com/actualbudget/actual/pull/727)

Step 1: expose a static list of the available banks (this PR)

Step 2: expose the full list of banks as per your suggestion (I'll update the existing PR with this additional logic: https://github.com/actualbudget/actual-server/pull/144)

fstybel (approved):

j-f1 (approved):

PR #147, opened 2023-03-06
Improve edge image build times

j-f1:


docker/edge-alpine.Dockerfile
```diff
@@ -1,25 +1,23 @@
...
-ADD "https://api.github.com/repos/actualbudget/actual/commits" /tmp/actual-commit.json
-RUN git clone --depth=1 https://github.com/actualbudget/actual /frontend
-RUN yarn install
-RUN ./bin/package-browser
+RUN mkdir /public
+ADD "https://api.github.com/repos/actualbudget/actual/actions/artifacts?name=actual-web&per_page=100" /tmp/artifacts.json
+RUN jq -r '[.artifacts[] | select(.workflow_run.head_branch == "jed/artifact")][0]' /tmp/artifacts.json > /tmp/latest-build.json
+
+ARG GITHUB_TOKEN
+RUN curl -L -o /tmp/desktop-client.zip --header "Authorization: Bearer ${GITHUB_TOKEN}" $(jq -r '.archive_download_url' /tmp/latest-build.json)
```
j-f1, line 13: It’s generally not a good idea to put secrets into Docker images since you can see all of the layers used to create the image, but in this case I think it’s fine because:

- The token is invalidated once the workflow completes (according to my reading)
- The image is not included in the final Docker image, so as far as I can tell there is no way to retrieve the token from the published image.

MatissJanis (approved):

Issue #151, opened 2023-03-08
[Bug]: Node 19 unsupported


0rnella:
Moving from where I had mistakenly filed it on https://github.com/actualbudget/actual/issues/726

TLDR: Sql-lite3 version ^7.5.0 is not compatible with node@19.

We could upgrade it -- it does look like [better-sqlite3](https://github.com/WiseLibs/better-sqlite3) on the latest version, 8.1.0, installs just fine with the latest node -- but unfortunately, there is no corresponding [@types/better-sqlite3](https://www.npmjs.com/package/@types/better-sqlite3) package released yet (I have opened [a discussion](https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/64626), though) and the latest one, 7.6.3, uses the incompatible version.

So the agreed solution in the meantime (discussed with @j-f1) could be to set the engines node version to be up to node@18 for easier debugging.

I say for easier debugging, because setting the engines version only gives a warning (or error, if used in conjunction with `engineStrict` if you are installing the package as a dependency) -- so this would be mostly for documentation purposes.

j-f1:
I took a look at the release notes for the 8.x releases and I don’t see any API changes so I think it could be fine to just update the `better-sqlite3` package for now and deal with the `@types` package later.

0rnella:
The `@types` packages has a dependency on the `better-sqlite3` package of the same version though, so that's not an option, unfortunately.
But we can also leave this open until I hear back re: updating the types package. https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/64626

PR #152, opened 2023-03-08
docs(engines): add node version in package.json engines

0rnella:
Per issue #151 , adding a node version to package.json engines!
This should be cleaned up once better-sqlite3 is upgraded to 8.x.x


j-f1:
The check failures will be fixed by https://github.com/actualbudget/actual-server/pull/153

MatissJanis:
Is this still relevant after v23.3.0 release? https://github.com/actualbudget/actual-server/pull/154

0rnella:
Hmm, keeping an older major version of a @types package around does sound a bit icky to me, since there's been breaking changes, but looking at the actual usage of it, I wonder if it's even needed at all? 
Anyway, yes, this release does make this PR unnecessary :)

MatissJanis:
> Hmm, keeping an older major version of a @types package around does sound a bit icky to me, since there's been breaking changes, but looking at the actual usage of it, I wonder if it's even needed at all? Anyway, yes, this release does make this PR unnecessary :)

Good point. We might consider cutting it :)

Anyways - sorry that we had to close this PR. But thanks for the help nonetheless! And thanks for adding node 19 support to better-sqlite3!

0rnella:
No worries, the PR was a band-aid, I'm happy that a better solution was found and it can be closed!

PR #153, opened 2023-03-08
Fix Docker actions failing on PRs from forks

j-f1:


.github/workflows/docker-edge.yml
```diff
@@ -56,12 +56,14 @@ jobs:
 
       - name: Login to Docker Hub
         uses: docker/login-action@v2
+        if: github.event_name != 'pull_request'
```
0rnella, line 59: chiming in to get some context 😅 

It sounds like on `pull_request`, we want to run the first 3 steps and ignore the last 3. What do the first 3 do? Would it be an option to just not this workflow at all by removing the trigger on l.11? 

And if we do want to run the first 3, have you considered separating this one job into 2, one job for the first 3 steps, and one for login+build+push that has this if condition?

j-f1: I want to build the container images to make sure that a PR doesn’t unintentionally break them, but they shouldn’t actually be pushed anywhere. (Which is why I’m disabling the two steps that log into the registries but not the actual container build itself)

I think we should keep the direct build action around as well since it gives faster feedback.

MatissJanis (approved):

PR #155, opened 2023-03-09
🔖 (23.3.1) fix nordigen usage in fly.io

MatissJanis:
This _slightly_ changes the interface for configuring Nordigen, but IMO it's safe to do because

1. the nordigen support is still experimental and subject to breaking changes
2. we released 23.3.0 only an hour ago

### DIFF
--- a/src/app-nordigen/services/nordigen-service.js
+++ b/src/app-nordigen/services/nordigen-service.js
@@ -17,8 +17,8 @@ import config from '../../load-config.js';
 
 const NordigenClient = nordigenNode.default;
 const nordigenClient = new NordigenClient({
-  secretId: config.nordigen_secret_id,
-  secretKey: config.nordigen_secret_key,
+  secretId: config.nordigen.secretId,
+  secretKey: config.nordigen.secretKey,
 });
 
 export const handleNordigenError = (response) => {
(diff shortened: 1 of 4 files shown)


j-f1 (commented):
Small nit but otherwise LGTM!

j-f1 (approved):

src/app-nordigen/services/nordigen-service.js
```diff
@@ -17,8 +17,8 @@ import config from '../../load-config.js';
 
 const NordigenClient = nordigenNode.default;
 const nordigenClient = new NordigenClient({
-  secretId: config.nordigen_secret_id,
-  secretKey: config.nordigen_secret_key,
+  secretId: config.nordigen.secretId,
+  secretKey: config.nordigen.secretKey,
 });
```
j-f1, line 22: This should probably not be set up at all if the config is not provided.

MatissJanis: I agree. Something we should fix as a follow-up. I'll add it to the list: https://github.com/actualbudget/actual/issues/724

j-f1: (Fine to defer to a future PR tho)

j-f1 (approved):

PR #157, opened 2023-03-10
Don’t make Dockerfile a symlink

j-f1:
People have been having issues with this. I copied it from `docker/stable-ubuntu.Dockerfile`. We should follow up on this to remove the duplication at some point but this will help fix the problems people are having now.


MatissJanis (approved):
Rubber stamp approval as I'm sadly no Docker expert. But from a high-level looks good :+1:

j-f1:
Thanks!

PR #160, opened 2023-03-11
:bug: (nordigen) close window when opening /nordigen/link path

MatissJanis:
When Nordigen accounts are linked 2x windows get opened: 1x (initial tab) is polling for the status; the other window is for authenticating with bank. Once the auth is done - customer is redirected back to app: `/nordigen/link` page. At this point we can just close the tab.

https://github.com/actualbudget/actual/issues/724#issuecomment-1462853590

### DIFF
--- a/src/app-nordigen/app-nordigen.js
+++ b/src/app-nordigen/app-nordigen.js
@@ -8,7 +8,6 @@ import {
 } from './errors.js';
 import { handleError } from './util/handle-error.js';
 import validateUser from '../util/validate-user.js';
-import config from '../load-config.js';
 
 const app = express();
 app.get('/link', function (req, res) {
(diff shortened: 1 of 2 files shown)


src/app-nordigen/app-nordigen.js
```diff
@@ -10,6 +10,18 @@ import { handleError } from './util/handle-error.js';
 import validateUser from '../util/validate-user.js';
 
 const app = express();
+app.get('/link', function (req, res) {
+  res.send(`
+<script>window.close();</script>
```
j-f1, line 15: Small nit: Ideally this would be a valid HTML document (including `<!DOCTYPE html>` so we don’t end up with quirks mode weirdness).

MatissJanis: Makes sense: https://github.com/actualbudget/actual-server/pull/160/commits/eb528cacfcd11f6287eb41c8ebc27b968d0e5d74

j-f1 (approved):
Nice!

PR #162, opened 2023-03-11
:sparkles: (nordigen) add status endpoint for checking status

MatissJanis:
Add a small endpoint that will be used to check the configuration status of Nordigen. If it is not configured - we should not allow people to proceed in the account-link flow.

### DIFF
--- a/src/app-nordigen/app-nordigen.js
+++ b/src/app-nordigen/app-nordigen.js
@@ -25,6 +25,15 @@ app.use(async (req, res, next) => {
   next();
 });
 
+app.post('/status', async (req, res) => {
+  res.send({
+    status: 'ok',
+    data: {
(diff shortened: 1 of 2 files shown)


j-f1 (approved):
Neat! Do you want to do a reachability test of the Nordigen API too so downtime can be diagnosed?

MatissJanis:
oo! That's a good idea. I'll see how difficult it would be to add that too.

MatissJanis:
Yea, so Nordigen doesn't provide a health-check endpoint. Theoretically we could generate a new token to perform this status check.. but I wonder if that will cause some other unintended consequences..

Given this: lets go ahead with the simple approach. If people start complaining that this is not sufficient - we can revisit the problem.

PR #167, opened 2023-03-18
Report parse issues for config.json in default location

Jackenmen:
This changes the behavior of configuration loading when ACTUAL_CONFIG_PATH env var is not specified. With this change, syntax errors in config.json will now be reported if ACTUAL_CONFIG_PATH env var is not specified and the app will not proceed. When the config.json file is not present or cannot be read, the behavior remains the same.


Jackenmen:
Does this repository not require me to add a release note file somewhere?

j-f1 (approved):
Looks good to me!

MatissJanis (approved):

MatissJanis:
Thanks!

PR #172, opened 2023-03-27
Stream file without loading to memory in /download-user-file

Jackenmen:
Briefly mentioned by @j-f1 here: https://discord.com/channels/937901803608096828/940290142579605514/1089653624390561822

I opted to keep the `500` HTTP code as is and therefore had to add a callback to `res.sendFile()` call and use a different code. I can instead update the test to check for `404` and remove the callback entirely.

This could be extended to `/upload-user-file` by removing the `application/actual-sync` and `application/encrypted-file` middleware and just use `req` as a stream by piping it into a file (with some added check for `Content-Length` header to ensure that we still have some max size for these files)

### DIFF
--- a/src/app-sync.js
+++ b/src/app-sync.js
@@ -291,7 +291,7 @@ app.post('/upload-user-file', async (req, res) => {
   }
 });
 
-app.get('/download-user-file', async (req, res) => {
+app.get('/download-user-file', async (req, res, next) => {
   let user = validateUser(req, res);
   if (!user) {
     return;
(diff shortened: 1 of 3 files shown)


Jackenmen:
`yarn build` is failing and I'm not sure how to fix it - the type should be narrowed down to `HttpError` by `instanceof` yet it isn't...

j-f1:
Thanks for opening this PR. Looks fine to me but I’d like @MatissJanis to take a look too in case I missed something silly.

> This could be extended to `/upload-user-file` by removing the `application/actual-sync` and `application/encrypted-file` middleware and just use req as a stream by piping it into a file (with some added check for `Content-Length` header to ensure that we still have some max size for these files)

Cool idea! I think this would also need to be implemented as a transform stream so that a malicious/broken client doesn’t accidentally upload longer content than the length header claims.

src/app-sync.js
```diff
@@ -313,17 +314,17 @@ app.get('/download-user-file', async (req, res) => {
...
+  res.sendFile(getPathForUserFile(fileId), function (err) {
+    if (err) {
+      if (err instanceof HttpError && err.status == 404) {
+        console.log('File does not exist on server.');
+        res.status(500).send('File does not exist on server.');
+      } else {
+        next(err);
+      }
+    }
+  });
```
MatissJanis, line 327: I think it's totally fine to return 404 in this case. Makes the code easier :) 

Just don't remember to update the test file if you do this.

```suggestion
  res.sendFile(getPathForUserFile(fileId));
```

Jackenmen: Welp, that also resolves my problem with how to get type narrowing to work so that's nice :smile:

Jackenmen:
I added a release note, I didn't notice I missed it.

> Cool idea! I think this would also need to be implemented as a transform stream so that a malicious/broken client doesn’t accidentally upload longer content than the length header claims.

I can work on this in the coming days. I will have to look into how transform streams work first since I don't know much about NodeJS so I can't spare time for it today but perhaps tomorrow. I somewhat assumed that the web framework would already fail once we go far enough into the stream to know that the Content-Length header is malformed but I assume it doesn't from what you're saying.

MatissJanis (approved):
Thanks for your contribution!

PR #174, opened 2023-03-28
:white_check_mark:  rename .test.js file to .spec.js

MatissJanis:
The test was not getting executed anymore because of the file extension.

Renaming it so it would be executed again.


MatissJanis:
This was already fixed: https://github.com/actualbudget/actual-server/pull/171

🤦

PR #185, opened 2023-04-09
Use babel preset-typescript

albertogasparin:
Seems like we already added `tsc` to build the project, but we use the wrong babel preset (Flow) instead of the specific TS one. This is only used in testing to make Jest work (from what I can tell).


j-f1 (approved):
🙅 no go flow

PR #186, opened 2023-04-10
Add a password set/reset script

j-f1:
This allows a couple of new workflows:

- setting the password before deploying the server, so that you don’t have to worry about an attacker bootstrapping it before you can
  ```shellsession
  $ npm run reset-password
  It looks like you don’t have a password set yet. Let’s set one up now!
  Enter a password, then press enter: hunter2
  Enter the password again, then press enter: hunter2
  Password set!
  ```
- resetting the password without having to delete the server-files database (which includes the file list we would like to not have to delete!
  ```shellsession
  $ npm run reset-password
  It looks like you already have a password set. Let’s reset it!
  Enter a password, then press enter: hunter2
  Enter the password again, then press enter: hunter2
  Password changed!
  ```

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "build": "tsc",
     "test": "NODE_ENV=test NODE_OPTIONS='--experimental-vm-modules --trace-warnings' jest  --coverage",
     "types": "tsc --noEmit --incremental",
-    "verify": "yarn -s lint && yarn types"
+    "verify": "yarn -s lint && yarn types",
+    "reset-password": "node src/scripts/reset-password.js"
   },
   "dependencies": {
     "@actual-app/api": "5.1.0",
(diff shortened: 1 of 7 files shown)


j-f1:
Fixing CodeQL error in #187.

j-f1:
I’ll look into ways of obscuring the input (such as with `****`). I don’t want to do the UNIX-style approach of hiding the password entirely since I’m worried people who aren’t used to that will think that something is broken.

j-f1:
Done!

MatissJanis (approved):
Very cool

BloodyHellcat:
Sorry for that low knowledge level question.
But how do I use this command on a docker instance of Actual Budget?

Thanks in advance for your response.

PR #189, opened 2023-04-16
Add a handler for AxiosError errors

j-f1:
This will help debug more clearly when Nordigen’s API returns unexpected status codes, since the current shape causes the actual response data to be collapsed in the console logs.

### DIFF
--- a/src/app-nordigen/app-nordigen.js
+++ b/src/app-nordigen/app-nordigen.js
@@ -1,6 +1,7 @@
 import { isAxiosError } from 'axios';
 import express from 'express';
 import path from 'path';
+import { inspect } from 'util';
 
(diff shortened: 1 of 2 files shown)


src/app-nordigen/app-nordigen.js
```diff
@@ -199,6 +200,16 @@ app.post(
             error_code: 'NORDIGEN_ERROR',
           });
           break;
+        case isAxiosError(error):
```
j-f1, line 203: Fun fact! This doesn’t actually narrow the `error` type away from `any`. Maybe https://github.com/microsoft/TypeScript/pull/53681 will fix that?

MatissJanis (approved):

PR #190, opened 2023-04-20
Add a field with all transactions to /nordigen/transactions endpoint

Jackenmen:
Helps with actualbudget/actual#919 by adding the `all` field wit both pending and booked transactions to the output of `getTransactionsWithBalance()` and, by extension, the `/nordigen/transactions` endpoint.

I could alter the `getTransactions()` to return the `all` field as well but I figured that keeping it such that it returns the output from Nordigen API 1:1 might be better so I left it as is. If you don't agree, let me know and I'll update this.

### DIFF
--- a/src/app-nordigen/services/nordigen-service.js
+++ b/src/app-nordigen/services/nordigen-service.js
@@ -170,7 +170,7 @@ export const nordigenService = {
    * @throws {RateLimitError}
    * @throws {UnknownError}
    * @throws {ServiceError}
-   * @returns {Promise<{iban: string, balances: Array<import('../nordigen-node.types.js').Balance>, institutionId: string, transactions: {booked: Array<import('../nordigen-node.types.js').Transaction>, pending: Array<import('../nordigen-node.types.js').Transaction>}, startingBalance: number}>}
+   * @returns {Promise<{iban: string, balances: Array<import('../nordigen-node.types.js').Balance>, institutionId: string, transactions: {booked: Array<import('../nordigen-node.types.js').Transaction>, pending: Array<import('../nordigen-node.types.js').Transaction>, all: Array<import('../nordigen.types.js').TransactionWithBookedStatus>}, startingBalance: number}>}
    */
   getTransactionsWithBalance: async (
     requisitionId,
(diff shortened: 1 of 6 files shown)


src/app-nordigen/services/nordigen-service.js
```diff
@@ -202,6 +202,13 @@ export const nordigenService = {
     const sortedPendingTransactions = bank.sortTransactions(
       transactions.transactions?.pending,
     );
+    const allTransactions = sortedBookedTransactions.map((t) => {
+      return { ...t, booked: true };
+    });
+    sortedPendingTransactions.forEach((t) =>
+      allTransactions.push({ ...t, booked: false }),
+    );
```
MatissJanis, line 210: **🥜 nitpick:** ‏
```suggestion
    const allTransactions = [
      ...transactions.transactions?.booked.map((t) => {
        return { ...t, booked: true };
      }),
      ...transactions.transactions?.pending.map((t) => {
        return { ...t, booked: false };
      }),
    ];
```

MatissJanis:
Sweet!

src/app-nordigen/services/nordigen-service.js
Jackenmen: I did it like this mostly because my way prevents unnecessary double iteration over the array but I imagine it's not a big deal :smile:

MatissJanis: Both solutions have double iteration btw.

But yea.. this is mostly just a nitpick :)

Jackenmen: I'll update this but could you tell me if there's any reason to prefer `transactions.transactions?.booked` over `sortedBookedTransactions` here?

MatissJanis: oh, you meant the spread operation. Yea, that's sub-optimal in this approach.

MatissJanis: > I'll update this but could you tell me if there's any reason to prefer transactions.transactions?.booked over sortedBookedTransactions here?

I imagine we'll want to remove `sortedBookedTransactions` a few releases later in favor of the new `all`. So it will make the future migration a bit easier. But again: no biggie. It's just a small nitpick. :)

Jackenmen: > oh, you meant the spread operation. Yea, that's sub-optimal in this approach.

Yep. I opted for a more performant way first and left it up to you to either tell me to use the cleaner way or the optimal way. But to be honest, optimizing this is a bit nitpicky on my side since I don't imagine it could possibly have any measurable difference with the number of transactions we could get at most.

henrikmaa:
The all-transactions are not being sent in the /transactions response

Need to add "all" to transactions here
https://github.com/actualbudget/actual-server/blob/master/src/app-nordigen/app-nordigen.js#L143&#L175

Jackenmen:
You're absolutely right, I don't know how I missed it, thanks!

MatissJanis (approved):
Thanks!

PR #194, opened 2023-04-24
App secret service

henrikmaa:
Adds support for secrets in actual-server

usage:

```
import { secretsService } from './services/secrets-service.js';

secretsService.set('name', 'value')
secretsService.get('name') //returns secret value or null if does not exist
secretsService.exists('name') //boolean
```

secretsService.get() should only be used inside actual-server and not sent to the client.

Endpoints for setting or checking if a secret exist in client

`POST /secret`
payload: `{name: 'secret', value: 'somevalue'}`
`200 - secret set`
`401`- unauthorized

`GET /secret/{name}`
`204 No Content` - if key exists
`404` - if does not exist
`401`- unauthorized

### DIFF
--- a/src/services/secrets-service.js
+++ b/src/services/secrets-service.js
@@ -2,7 +2,7 @@ import createDebug from 'debug';
 import fs from 'node:fs';
 import config, { sqlDir } from '../load-config.js';
 import { join } from 'node:path';
-import openDatabase from '../db.js';
+import getAccountDb from '../account-db.js';
 
 /**
  * An enum of valid secret names.
@@ -18,6 +18,7 @@ class SecretsDb {
   constructor() {
     this.debug = createDebug('actual:secrets-db');
     this.db = null;
+    this.initialize();
     this.migrateNordigen();
   }
 
@@ -36,26 +37,19 @@ class SecretsDb {
     }
   }
 
+  initialize() {
+    if (!this.db) {
+      this.db = this.open();
+    }
+
+    this.debug(`initializing secrets table'`);
+    //Create secret table if it doesn't exist
+    const initSql = fs.readFileSync(join(sqlDir, 'secrets.sql'), 'utf8');
+    this.db.exec(initSql);
+  }
+
   open() {
-    if (!fs.existsSync(config.serverFiles)) {
-      this.debug(`creating server files directory: '${config.serverFiles}'`);
-      fs.mkdirSync(config.serverFiles);
-    }
-
-    const dbPath = join(config.serverFiles, 'secrets.sqlite');
-    const needsInit = !fs.existsSync(dbPath);
-
-    let db = openDatabase(dbPath);
-
-    if (needsInit) {
-      this.debug(`initializing secrets database: '${dbPath}'`);
-      let initSql = fs.readFileSync(join(sqlDir, 'secrets.sql'), 'utf8');
-      db.exec(initSql);
-    } else {
-      this.debug(`opening secrets database: '${dbPath}'`);
-    }
-
-    return db;
+    return getAccountDb();
   }
 
   set(name, value) {
--- a/src/sql/secrets.sql
+++ b/src/sql/secrets.sql
@@ -1,4 +1,4 @@
-CREATE TABLE secrets (
+CREATE TABLE IF NOT EXISTS secrets (
   name TEXT PRIMARY KEY,
   value BLOB
 );


henrikmaa:
@MatissJanis Any toughts on how to implement it for the nordigen service? should we keep the current `config.nordigen` as a fallback?

MatissJanis:
> @MatissJanis Any toughts on how to implement it for the nordigen service? should we keep the current `config.nordigen` as a fallback?

Can we create a migration that picks up `config.nordigen` and moves it to the new approach? I'd like to eventually sunset the nordigen key usage via env vars/config json.

src/services/secrets-service.js
```diff
@@ -0,0 +1,105 @@
...
+  get(name) {
+    if (!this.db) {
+      this.db = this.open();
+    }
+
+    this.debug(`getting secret '${name}'`);
+    const result = this.db.first(`SELECT value FROM secrets WHERE name =?`, [
+      name,
+    ]);
+    this.close();
```
j-f1, line 81: Why close the DB at the end of every get/set?

src/services/secrets-service.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+const secretsDb = new SecretsDb();
+export const secretsService = {
+  get: (name) => {
+    const secret = secretsDb.get(name)?.value;
+    if (!secret) return null;
+
+    return secretsDb.get(name)?.value;
```
j-f1, line 97: No need to call `get()` twice, right?

```suggestion
    return secretsDb.get(name)?.value ?? null;
```

src/app-nordigen/services/nordigen-service.js
```diff
@@ -15,11 +15,14 @@ import * as nordigenNode from 'nordigen-node';
...
 import config from '../../load-config.js';
 import jwt from 'jws';
+import { secretsService } from '../../services/secrets-service.js';
 
 const NordigenClient = nordigenNode.default;
 const nordigenClient = new NordigenClient({
-  secretId: config.nordigen?.secretId,
-  secretKey: config.nordigen?.secretKey,
+  secretId:
+    secretsService.get('nordigen_secretId') || config.nordigen?.secretId,
```
j-f1, line 23: It feels like it would be super easy to typo these secret names. What do you think about defining a `secretNames` object with all valid secret names so that you could do `secretsService.get(secretNames.nordigen.secretId)`?

j-f1 (commented):
Thank you for working on this! I like the overall approach but have some comments about the details:

src/services/secrets-service.js
henrikmaa: I would guess it would spare some memory usage, since it's probably that will not get accessed to often. But happy to remove it if you dont think it makes any difference

src/app-nordigen/services/nordigen-service.js
```diff
@@ -15,11 +15,14 @@ import * as nordigenNode from 'nordigen-node';
...
+  secretKey:
+    secretsService.get('nordigen_secretKey') || config.nordigen?.secretKey,
```
MatissJanis, line 25: **💬 suggestion:** ‏
```suggestion
  secretId:
    secretsService.get('nordigen_secretId'),
  secretKey:
    secretsService.get('nordigen_secretKey'),
```

src/services/secrets-service.js
```diff
@@ -0,0 +1,105 @@
...
+import openDatabase from '../db.js';
+class SecretsDb {
+  constructor() {
+    this.debug = createDebug('actual:secrets-db');
+    this.db = null;
+    this.migrateNordigen();
+  }
+
+  /// Migrates nordigen from config.json or process.env to app secret
+  migrateNordigen() {
```
MatissJanis, line 14: **💬 suggestion:** ‏instead of writing the migration in the codebase - could you use the `migrations`? Example here: https://github.com/actualbudget/actual-server/blob/master/migrations/1632571489012_remove_cache.js

src/services/secrets-service.js
```diff
@@ -0,0 +1,105 @@
...
+}
+
+const secretsDb = new SecretsDb();
+export const secretsService = {
+  get: (name) => {
+    const secret = secretsDb.get(name)?.value;
+    if (!secret) return null;
+
+    return secretsDb.get(name)?.value;
+  },
```
MatissJanis, line 98: **🔨 warning:** ‏this code runs 2x SQL queries. Could we instead call `get()` only 1x?

src/sql/secrets.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE secrets (
```
MatissJanis, line 1: **💭 thought:** ‏I wonder if we should use a brand-new database for this. Or perhaps re-use the existing `account`. What do you think @j-f1 ?

MatissJanis (changes requested):
Overall looking really good!

src/app-nordigen/services/nordigen-service.js
MatissJanis: Ohhh... one more thing here to check: are we calling an SQL query to get these values even if the user is not using nordigen? Say: if they are making a non-nordigen API call? Does the class still get instatiated and thus makes the SQL calls? If it does - we need to figure out a way to fix it.

henrikmaa: The secret service does not execute any sql calls until either .get(), .set() or .exists() are executed

src/services/secrets-service.js
henrikmaa: is it ok if i use the stuff from the secretService inside the nordigen_secret.js migration file then? or should i do it all "manually"

src/app-nordigen/services/nordigen-service.js
MatissJanis: Right, but you're just building a JSON object here with the secret values. Thus the queries would be executed as long as the whole file gets loaded and executed.

A bit tricky to explain.. But basically there is no difference between these two usages.. in both cases the underlying SQL gets executed:

```
// 1:
const hello = { id: secretSecrvive.get("secret") };

// 2:
const nordigen = new Nordigen({ id: secretSecrvive.get("secret") })
```

src/services/secrets-service.js
MatissJanis: Preferably manually. The shape & logic in this code will eventually evolve into something else. But the migration should stay exactly the same throughout its history.

src/app-nordigen/services/nordigen-service.js
henrikmaa: Ah yes, i could create a Map  to cache values?

something like this?

```js
export const secretsService = () => {
  const _cachedSecrets = new Map();
 
  const get = (name) => {
    return _cachedSecrets.get(name) ?? secretsDb.get(name)?.value ?? null;
  };

  const set = (name, value) => {
    const { changes } = secretsDb.set(name, value);
    if (changes) {
      _cachedSecrets.set(name, value);
    }
    return changes;
  };

  const exists = (name) => {
    return Boolean(get(name));
  };
  
  return { get, set, exists };
};
```

henrikmaa:
I will fix the migration tomorrow :) Thanks for the suggestions!

src/app-nordigen/services/nordigen-service.js
MatissJanis: What happens in this scenario?

1. customer does not have secrets
2. customer attempt to call nordigen (but this obviously fails)
3. customers sets the new credentials
4. customer attempts to call nordigen again

Does the 2nd nordigen call retrieve the new credentials? Or the cached?

--

Just tested: the nordigen keys are currently retrieved only 1x upon server start. Which means in step (4) the old keys would be retrieved. This means we'll need to instatiate a new nordigen service class each time we call the nordigen `actual-server` APIs. Which means a bit of refactors.. sadly 😢

henrikmaa: Yeah I guess we would have this issue for any solution we would go for regarding updating keys from frontend.

The only workaround solution I got is to create an onUpdate event that can be listened to in the nordigen class. Then we do some logic there to make sure the values update. 

Then the nordigen class should be able to update its own secrets without reinitialising

src/services/secrets-service.js
henrikmaa: I cant figure out how to do the migration. Appreciate some help on this.

j-f1: Those are the migrations for the frontend database, I think we should remove them from this repo altogether. Since we don’t have migrations set up at the moment I don’t think it makes sense to set those up, we can just keep this chunk of code in here for ~6 months or so.

src/sql/secrets.sql
j-f1: Hm… I think since there are already a bunch of things in the account db, I think we should add this table there instead of as a separate DB.

src/services/secrets-service.js
MatissJanis: Good point. I actually thought the migrations in `actual-server` were working.. but it turns out they don't 🤔 

So yes: lets keep your existing migration solution. Sorry for the noise.

henrikmaa:
I think all the review suggestions have been solved. Let me know if the `secretsService.onUpdate` is something we dont want. Otherwise it should all be good to go :)

Thanks for taking the time!

src/app-nordigen/services/nordigen-service.js
```diff
@@ -13,13 +13,20 @@ import {
...
+  secretId: secretsService.get(SecretName.nordigen_secretId),
+  secretKey: secretsService.get(SecretName.nordigen_secretKey),
+});
+
+secretsService.onUpdate(SecretName.nordigen_secretId, (newSecret) => {
+  nordigenClient.secretId = newSecret;
+});
+secretsService.onUpdate(SecretName.nordigen_secretKey, (newSecret) => {
+  nordigenClient.secretKey = newSecret;
 });
```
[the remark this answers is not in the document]
henrikmaa: Im split on this. In one way the logic is prettier to look at for the secretsService code. But it also feels a bit wrong always reinitialising.

Overall other classes in the future that wants to use secrets, also need to always reinitialise then to make sure they get latest keys.

Not sure whats the best tradeoff

MatissJanis (approved):

src/app-nordigen/services/nordigen-service.js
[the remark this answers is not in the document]
MatissJanis: Tricky. Lets go with your solution now. We can always re-visit this and change it if we figure out a better solution.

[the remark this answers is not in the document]
henrikmaa: I tried to test this

```
function getNordigenClient() {
  return new NordigenClient({
    secretId: secretsService.get(SecretName.nordigen_secretId),
    secretKey: secretsService.get(SecretName.nordigen_secretKey),
  });
}
```

usage example:

```
isConfigured: () => {
    return !!(getNordigenClient().secretId && getNordigenClient().secretKey);
  },
```

But this did not work. It still used the old keys. Not sure why
Even if we dont go for this solution now, dont you think this should work?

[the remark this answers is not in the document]
MatissJanis: That is strange.. Yes, it should have initialised a new class and gotten new secrets. I wonder what's going on there 🤔

PR #195, opened 2023-04-25
Remove the unused `migrations` folder

j-f1:
As far as I can tell, nothing in the codebase references it. If we ever need it for some reason, it is bundled with the frontend.


MatissJanis (approved):
thanks! I wonder why we even had these here

j-f1:
Probably for the full sync backend.

PR #210, opened 2023-06-03
Thin translation layer for a RESTful Actual API

anh-chu:
category: Features
authors: [anh-chu]
Add a RESTful APIs that translates requests to the node.js module's functions
Add new config options related to the APIs
Add new infrastructure requirement (redis) to control lock-based queue for API requests

NOTE: this has not been integrated into the authentication, session, error handling, and config mechanism of the main code base. WIP.


anh-chu:
This is a stateful API. It's working decently at the moment, except for the things noted above (auth. session, and error handling use different mechanisms). There are 3 main endpoints:
- `/api/init`: this takes the budget ID and password (if encrypted) to authenticate the user with the server & get the budget file. This needs to be called first for any client.
- `/api/query`: this translate a JSON request body to the ActualQL functions. For example:
`{
    "q": "transactions",
    "select": "date",
    "orderBy": {
        "date": "desc"
    }
}`
- `/api/:function`: `:function` is the nodejs package's functions such as `getTransactions` or `getAccounts`. Arguments are passed in the request body. However, the design here may be changed slightly to accommodate for a better syntax.

To avoid any potential issues when multiple clients are interacting with the endpoints or interacting with multiple budget files, I have implemented a basic lock-based queue with Redis. This, when combined with sessions, should isolate clients decently.

anh-chu:
I have decided to not adapt the authentication and session of the main code at the moment since that mechanism cannot accommodate securely for different devices at the moment. My addition is relatively small, so I'll stick with it and make it an optional deployment option for people who wants to have to API. If people do not want to enable the API, they would not have to provision a Redis server.

anh-chu:
Biggest change I introduced is probably the gradual transition to Typescript. This allows for a lot more flexibility while adhering to good typing practices.
The mechanism for API auth is different, using JWT token stored in Redis to isolate clients and credentials. The JWT is expected in Authorization header.
The next enhancement for this is error handling, but it's ready for use at the moment.

anh-chu:
Some stuff I noticed:
- Seems like you intended this to work with Node 18? Not sure, but I am using Node 16.
- I want to do more configuration for this to work with pm2 after build for the sake of optimization
-

MatissJanis:
:wave: Hey! Thanks for the contribution. It's always amazing to see what the community can come up with!

At this point Actual is using a pull-based architecture. Adding a RESTful API would contradict this approach and move the primary database away from the client to the server (thus in-part negating the local-first aspect of Actual). Which is why currently we're not looking to add RESTful API capabilities to the sync server. But again - thanks for bringing this idea to the table!

And apologies for taking so long to review.

Issue #203, opened 2023-05-12
Activating HTTPS kills HTTP


P1514:
Hello,

I'm currently trying to deploy this on truenas (docker)

Without HTTPS it asks for Shared....

When I configure HTTPS, it works (locally on the docker) with one catch. HTTP requests don't get forwared/upgraded to HTTPS. Just empty reply from server

This is an issue, since truenas detect the app is up by calling http://XXXX:5006, and since HTTP fails the app is never up so forwarding is never active.

Anyone else experiencing this issue?

j-f1:
I don’t know if there is a way to configure Node.js’s HTTPS module to do that redirect, but if you can figure it out it would be much appreciated!

P1514:
Nodejs is definetly not my confort language.
Quick search I think the best way would be to have two listeners http on 5006 and https on 5007, http would forward to https if it is configured.

I'll try to do something on my end see if it would work

j-f1:
I don’t know if we would want to do that since it would break containers for people who already have them set up to use HTTPS. There seems to be a package `httppolyglot` that does this but it’s old and unmaintained.

AdrianAcala:
@P1514 , I threw in a fix. Please try my PR and let me know if that fixes it.

If it doesn't, please share how to recreate the issue and I'll try again.

kamarkiewicz:
Hi, I would like to suggest a different approach if I may. You can add a reverse proxy like nginx to that docker image. It can handle both protocols on one port, redirect HTTP to HTTPS, and work well with services like TrueNAS that does HTTP checks. Sample nginx config:

```nginx
server {
    listen 5006 ssl http2;
    listen 5006;

    server_name localhost;

    ssl_certificate /path/to/your/server.cert;
    ssl_certificate_key /path/to/your/server.key;

    # Redirect all HTTP traffic to HTTPS
    if ($scheme = http) {
        return 301 https://$server_name$request_uri;
    }

    location / {
        proxy_pass http://localhost:3000; # Actual internal port here
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
    }
}

AdrianAcala:
[...]

PR #211, opened 2023-06-04
WIP #203 Fixing issue with mixed content

AdrianAcala:
This is to fix issue https://github.com/actualbudget/actual-server/issues/203.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -11,21 +11,28 @@ import * as syncApp from './app-sync.js';
 import * as nordigenApp from './app-nordigen/app-nordigen.js';
 import * as secretApp from './app-secrets.js';
 
-const app = express();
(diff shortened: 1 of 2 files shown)


j-f1:
It looks like you’re now opening port 80. I don‘t think this will work if you’re running on a machine that’s already listening on port 80 for something unrelated, and with a container, only port 5006 would continue to be exposed. Additionally, `http://<actual-ip>:5006` would continue to error out since that port only listens for HTTPS connections.

AdrianAcala:
@j-f1 , if you have something else listening on port 80, like Nginx, you should redirect or reverse proxy using the web server. You can't expect this application to handle that redirect for you.

Closing this PR as it is no longer necessary.

j-f1:
Thanks for giving this fix a shot!

PR #212, opened 2023-06-06
Add missing step to deploy to README

enxyz:
Fly now requires an additional deploy step before opening the URL. The documentation is very smooth for a new user, and having this called out will help stay in the fly. It is obvious what to do from the fly docs so this should be considered quite minor.

`flyctl launch` results in `✓ Configuration is valid` and the msg `Now: run 'fly deploy' to deploy your Node app.`. Trying `flyctl open` at this point will open the URL but there will be no app there yet. So, one must `fly deploy` in order to _have a running app_


MatissJanis:
👋 Could you please move the release note under `212.md` instead of the README.md file?

Thanks!

enxyz:
Done! Sorry, must have read the docs wrong

j-f1 (approved):
Thanks!

enxyz:
Thanks for the fix!

rich-howell:
I would like to keep all of the deployment documentation in one place, I have opened #215 which removes the bulk of the deployment documentation from the readme in favour of links to the documentation, thank you so much for your contribution @enxyz I will take these across to the docs site and check to make sure they are consistent there.

PR #218, opened 2023-06-18
:recycle: (crdt) use actual-app/crdt package instead of api

MatissJanis:
Using the new CRDT package instead of API.


j-f1 (approved):
Sweet! This should increase image sizes too.

PR #231, opened 2023-07-19
:recycle:  rename nordigen to gocardless

MatissJanis:
Just renaming things from Nordigen to Gocardless.


src/app-gocardless/README.md
```diff
@@ -3,7 +3,7 @@
 Find in [doc](https://docs.google.com/spreadsheets/d/1ogpzydzotOltbssrc3IQ8rhBLlIZbQgm5QCiiNJrkyA/edit#gid=489769432) what is id of bank which you want to integrate
 
 Add the `institution_id` and your name to list of possible options in the frontend
-project `actual/packages/loot-design/src/components/modals/NordigenExternalMsg.js`
+project `actual/packages/loot-design/src/components/modals/GoCardlessExternalMsg.js`
```
j-f1, line 6: It’s not `loot-design` anymore

src/services/secrets-service.js
```diff
@@ -19,11 +19,11 @@ class SecretsDb {
...
+    this.migrateGoCardless();
   }
 
-  /// Migrates nordigen from config.json or process.env to app secret
-  migrateNordigen() {
+  /// Migrates GoCardless from config.json or process.env to app secret
+  migrateGoCardless() {
     if (
       config.nordigen &&
       (!this.get(SecretName.nordigen_secretId) ||
```
j-f1, line 29: What do you want to do about the secret names? Would it make sense to migrate them to `gocardless_*`?

MatissJanis: Yea, I'd like to do this, but separately. It's a non-trivial operation to do this with backwards compatibility both in the UI and API layer.

src/app-gocardless/README.md
MatissJanis: Good catch. https://github.com/actualbudget/actual-server/pull/231/commits/39522714a741bf34e6907c823e04b25200421b8f

j-f1 (approved):

PR #237, opened 2023-08-01
Add NORWEGIAN-XX-NORWNOK1 GoCardless integration

kyrias:
This adds explicit support for all branchs of Bank Norwegian to the GoCardless integration to work around the issue reported in actualbudget/actual#1392 where their use of `valueDate` does not match the NextGenPSD2 definition which is what we expect to receive from GoCardless.  To work around this we simply remove this field when normalizing transactions and let the Actual app use `bookingDate` instead.

Resolves actualbudget/actual#1392.

### DIFF
--- a/src/app-gocardless/bank-factory.js
+++ b/src/app-gocardless/bank-factory.js
@@ -2,6 +2,7 @@ import AmericanExpressAesudef1 from './banks/american-express-aesudef1.js';
 import IngPlIngbplpw from './banks/ing-pl-ingbplpw.js';
 import IntegrationBank from './banks/integration-bank.js';
 import MbankRetailBrexplpw from './banks/mbank-retail-brexplpw.js';
+import NorwegianXxNorwnok1 from './banks/norwegian-xx-norwnok1.js';
 import SandboxfinanceSfin0000 from './banks/sandboxfinance-sfin0000.js';
 
 const banks = [
@@ -9,6 +10,7 @@ const banks = [
   IngPlIngbplpw,
   MbankRetailBrexplpw,
   SandboxfinanceSfin0000,
+  NorwegianXxNorwnok1,
 ];
 
 export default (institutionId) =>
--- a/src/app-gocardless/banks/american-express-aesudef1.js
+++ b/src/app-gocardless/banks/american-express-aesudef1.js
@@ -1,4 +1,4 @@
-import { amountToInteger, sortByBookingDate } from '../utils.js';
+import { amountToInteger, sortByBookingDateOrValueDate } from '../utils.js';
 
 /** @type {import('./bank.interface.js').IBank} */
 export default {
@@ -32,7 +32,7 @@ export default {
   },
 
   sortTransactions(transactions = []) {
-    return sortByBookingDate(transactions);
+    return sortByBookingDateOrValueDate(transactions);
   },
 
   /**
--- a/src/app-gocardless/banks/integration-bank.js
+++ b/src/app-gocardless/banks/integration-bank.js
@@ -1,4 +1,8 @@
-import { sortByBookingDate, amountToInteger, printIban } from '../utils.js';
+import {
+  sortByBookingDateOrValueDate,
+  amountToInteger,
+  printIban,
+} from '../utils.js';
 
 const SORTED_BALANCE_TYPE_LIST = [
   'closingBooked',
@@ -41,7 +45,7 @@ export default {
       'Available (first 10) transactions properties for new integration of institution in sortTransactions function',
       { top10Transactions: JSON.stringify(transactions.slice(0, 10)) },
     );
-    return sortByBookingDate(transactions);
+    return sortByBookingDateOrValueDate(transactions);
   },
 
   calculateStartingBalance(sortedTransactions = [], balances = []) {
--- a/src/app-gocardless/banks/norwegian-xx-norwnok1.js
+++ b/src/app-gocardless/banks/norwegian-xx-norwnok1.js
@@ -1,0 +1,100 @@
+import {
+  printIban,
+  amountToInteger,
+  sortByBookingDateOrValueDate,
+} from '../utils.js';
+
+/** @type {import('./bank.interface.js').IBank} */
+export default {
+  institutionIds: [
+    'NORWEGIAN_NO_NORWNOK1',
+    'NORWEGIAN_SE_NORWNOK1',
+    'NORWEGIAN_DE_NORWNOK1',
+    'NORWEGIAN_DK_NORWNOK1',
+    'NORWEGIAN_ES_NORWNOK1',
+    'NORWEGIAN_FI_NORWNOK1',
+  ],
+
+  normalizeAccount(account) {
+    return {
+      account_id: account.id,
+      institution: account.institution,
+      mask: account.iban.slice(-4),
+      iban: account.iban,
+      name: [account.name, printIban(account)].join(' '),
+      official_name: account.product,
+      type: 'checking',
+    };
+  },
+
+  normalizeTransaction(transaction, booked) {
+    /**
+     * The way Bank Norwegian handles the date fields is rather strange and
+     * countrary to GoCardless's documentation.
+     *
+     * For booked transactions Bank Norwegian sends a `valueDate` field that
+     * doesn't match the NextGenPSD2 definition of `valueDate` which is what we
+     * expect to receive from GoCardless.  Therefore we remove the incorrect
+     * field so that transactions are correctly imported.
+     */
+    if (booked) {
+      delete transaction.valueDate;
+      return transaction;
+    }
+
+    /**
+     * For pending transactions there are two possibilities.  Either the
+     * transaction has a `valueDate`, in which case the `valueDate` we receive
+     * corresponds to when the transaction actually occurred, and so we simply
+     * return the transaction as-is.
+     */
+    if (transaction.valueDate !== undefined) {
+      return transaction;
+    }
+
+    /**
+     * If the pending transaction didn't have a `valueDate` field then it
+     * should have a `remittanceInformationStructured` field which contains the
+     * date we expect to receive as the `valueDate`.  In this case we extract
+     * the date from that field and set it as `valueDate`.
+     */
+    if (transaction.remittanceInformationStructured) {
+      const remittanceInfoRegex = / (\d{4}-\d{2}-\d{2}) /;
+      const matches =
+        transaction.remittanceInformationStructured.match(remittanceInfoRegex);
+      if (matches) {
+        transaction.valueDate = matches[1];
+        return transaction;
+      }
+    }
+
+    /**
+     * If neither pending case is true we return `null` and ignore the
+     * transaction until it's been further processed by the bank.
+     */
+    return null;
+  },
+
+  sortTransactions(transactions = []) {
+    return sortByBookingDateOrValueDate(transactions);
+  },
+
+  /**
+   *  For NORWEGIAN_XX_NORWNOK1 we don't know what balance was
+   *  after each transaction so we have to calculate it by getting
+   *  current balance from the account and subtract all the transactions
+   *
+   *  As a current balance we use `expected` balance type because it
+   *  corresponds to the current running balance, whereas `interimAvailable`
+   *  holds the remaining credit limit.
+   */
+  calculateStartingBalance(sortedTransactions = [], balances = []) {
+    const currentBalance = balances.find(
+      (balance) => 'expected' === balance.balanceType,
+    );
+
+    return sortedTransactions.reduce((total, trans) => {
+      return total - amountToInteger(trans.transactionAmount.amount);
+    }, amountToInteger(currentBalance.balanceAmount.amount));
+  },
+};
--- a/src/app-gocardless/gocardless-node.types.ts
+++ b/src/app-gocardless/gocardless-node.types.ts
@@ -433,7 +433,7 @@ export type Transaction = {
   /**
    * Reference as contained in the structured remittance reference structure
    */
-  remittanceInformation?: string;
+  remittanceInformationStructured?: string;
 
   /**
    * The amount of the transaction as billed to the account
--- a/src/app-gocardless/tests/utils.spec.js
+++ b/src/app-gocardless/tests/utils.spec.js
@@ -1,5 +1,5 @@
 import { mockTransactionAmount } from '../services/tests/fixtures.js';
-import { sortByBookingDate } from '../utils.js';
+import { sortByBookingDateOrValueDate } from '../utils.js';
 
 describe('utils', () => {
   describe('#sortByBookingDate', () => {
@@ -18,7 +18,7 @@ describe('utils', () => {
           transactionAmount: mockTransactionAmount,
         },
       ];
-      expect(sortByBookingDate(transactions)).toEqual([
+      expect(sortByBookingDateOrValueDate(transactions)).toEqual([
         {
           bookingDate: '2023-01-20',
           transactionAmount: mockTransactionAmount,
--- a/src/app-gocardless/utils.js
+++ b/src/app-gocardless/utils.js
@@ -6,9 +6,11 @@ export const printIban = (account) => {
   }
 };
 
-export const sortByBookingDate = (transactions = []) =>
+export const sortByBookingDateOrValueDate = (transactions = []) =>
   transactions.sort(
-    (a, b) => +new Date(b.bookingDate) - +new Date(a.bookingDate),
+    (a, b) =>
+      +new Date(b.bookingDate || b.valueDate) -
+      +new Date(a.bookingDate || a.valueDate),
   );
 
 export const amountToInteger = (n) => Math.round(n * 100);
--- a/upcoming-release-notes/237.md
+++ b/upcoming-release-notes/237.md
@@ -1,0 +1,6 @@
+---
+category: Enhancements
+authors: [kyrias]
+---
+
+Add all integration for all Bank Norwegian branches to the GoCardless support.


Inrego:
Is this only for the Swedish branch? I am using the Danish branch

leikoilja (commented):
I am unfortunately not familiar of how the entire integration work, so it would be good for someone with prior knowledge to review as well.

@kyrias, i have just tried the new provider that you added and when adding a new card i get an error: 
image

i quickly placed a console debug right before normalising the transactions:
image

and can see that the shape of `response.transactions` contains both booked and pending transactions. It would be amazing to grab both of those sections, so maybe we need to concat those arrays?

For the experiment i tried just using `booked` as `response.transactions.booked.map` and i can verify that the starting balance has been calculated **correctly**, however there is no actual transactions imported:
image

leikoilja:
@Inrego, i would assume that Norwegian bank passes the same data structure to GoCardless independently of country (Sweden, Denmark or Norway), but there is no certain way to tell unless you try it with your Danish card to make sure :) if the custom added provider would work with your danish branch as well out of the box - i believe we can make the added provided generic and use for all Norwegian branches?

leikoilja:
@kyrias, checked-out the latest changes and right after adding a new "bank link" i get internal server error, no container error logs , but console error:
image

if i click 'sync' i get:
image

kyrias:
That's strange.  I wonder if there's a transaction without a `bookingDate`?  But it's listed as `cleared` in the error, hm.

kyrias:
Oh wait, this is the starting balance transaction that it's trying to add!  If you check in `https://bankaccountdata.gocardless.com/data/` and link the same card, do you have transactions under `pending` without a `bookingDate`?

I guess these should be sorted _after_ everything with a `bookingDate`.

kyrias:
If it is indeed the case that these were pending transactions, then I just pushed a commit which should solve it by _prepending_ rather than _appending_ the pending transactions to the booked ones.

leikoilja:
hmmm, still the same errors as above
image

and yes indeed, i have about 5 pending transactions with no `bookingData`, although they do have a `valueDate` that is the date when i tapped the card:
image

kyrias:
Ah, silly me, of course it'll fail with the same error but in a different place when trying to actually create the transactions.  :sweat_smile:

Their use of `valueDate` is so absurd, but if this is how they use it then that seems like a good solution.  I've pushed a new commit implementing that logic now.

leikoilja:
i m confused... still the same issue. would it help if i get the raw transactions from https://bankaccountdata.gocardless.com/data/ and grab the log transactions from actual logs as before? 🤔 
 
demo

kyrias:
This is super strange.  So it's again failing when creating the starting balance transaction (`starting_balance_flag: true`) and because the date isn't set.

To get this date the Actual app does the following:

https://github.com/actualbudget/actual/blob/907571bd83c105812e57c5421b4af7ef9035ff34/packages/loot-core/src/server/accounts/sync.ts#L817-L822

So _somehow_ we end up picking an `oldestTransaction` where `valueDate` _and_ `bookingDate` is unset.  But with the last changes we should only remove `valueDate` from booked transactions, so this is super confusing.

Would you be able to try running the app with an additional `console.log(oldestTransaction);`?  That way we should be able to figure out what's going on.

> would it help if i get the raw transactions from https://bankaccountdata.gocardless.com/data/ and grab the log transactions from actual logs as before? thinking

The current logs wouldn't be useful since it probably doesn't include the problematic transaction, but if you download the transaction log from the GoCardless portal you could try piping it through e.g. `jq '.transactions.booked[] | select(.valueDate == null)'` to see if you find any booked transactions without a value date.  Just logging the transaction above would probably be more useful though, and then find the raw data for that transaction in the portal by ID.

leikoilja:
@kyrias, any ideas how to debug where the error happens and the code explodes? 🤔 
i have set a console debugger right after we grab `olderstTransaction` but the code never seem to reach this stage 🤔 
image

kyrias:
Very confusing.

What if you instead go to the _Sources_ tab in the dev tools, then in the file tree go under the one start's called something like `browser-service`, then `localhost:3001 -> kcab -> kcab.worker.dev.js`.  In that file search for `function syncGoCardlessAccount` and scroll down to the `const oldestTransaction` line.  Then click on the line number which should both bring you to the reconstructed file from the source map _and_ set a breakpoint on that line.

If you then select the `browser-server` thread in the right sidebar and try to sync again then it should hopefully break on that line and let you see the local value.

leikoilja:
wow, awesome, thanks for the debugging tips, @kyrias. 

Okay, here is what i found:
image


what's more weird if that the last 5 transactions look like there:
image

seems like there might be pending transactions that have not been yet "finished processing" (weird weird).
Especially the last 2:
image

can't really find any logical explanation on those last 2 transactions and why they have no date at all. They are not even that difference from the transaction right about them...

What do you think if in `normalization` function we loop through transactions and transaction has no `bookedDate` or `valueDate` - we remove that transaction from the `transactions`. That would remove "unprocessed" transactions from Actual. And hopefully once they finish processing on GoCardLess/Norwegian side of things - Actual will pick them up after `sync` button is hit? 🤔

kyrias:
@leikoilja That is so weird!

One thing I noticed is that `remittanceInformationStructured` of those transactions have two datetimes, and the first of those matches the `valueDate` of the pending transaction that has one.  So if that's consistently the case then what we could do for pending transactions is:

- If it has a `valueDate` then retain it.
- Otherwise extract the first date from `remittanceInformationStructured` and if we found one set `valueDate` to it.
- If we can't parse a date out of `remittanceInformationStructured` we remove the transaction.

To do this I'd need to know the structure of that first blurred part of that field though.  I assume it's some kind of ID, which presumably doesn't contain spaces?  If so we can then just split on spaces.

kyrias:
@leikoilja I've pushed the changes I suggested above now, please see if it behaves as expected now.  :)

src/app-gocardless/banks/norwegian-se-norwnok1.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * If the pending transaction didn't have a `valueDate` field then it
+     * should have a `remittanceInformationStructured` field which contains the
+     * date we expect to receive as the `valueDate`.  In this case we extract
+     * the date from that field and set it as `valueDate`.
+     */
+    if (transaction.remittanceInformationStructured) {
+      const remittanceInfoRegex = /(?:[^ ]+ ){5}(\d{4}-\d{2}-\d{2})/;
```
leikoilja, line 55: Sorry for blurring the screenshot, but the real `remittanceInformationStructured` looks like this with the credit card number:
`111111******2222  SEK 39 SEK 39 2023-08-04 00:20:23 SL Stockholm SE    , 39, 2023-08-08T00:00:00.000`

so it seems we would be [safe enough to go with](https://regex101.com/r/CB7M6c/1)
```suggestion
      const remittanceInfoRegex = /(\d{4}-\d{2}-\d{2})/;
```

leikoilja (commented):
sorry, @kyrias, it's still error'ing. thanks for sticking on so far 💪 

The proposed regex expression does not cater for the full format of `remittanceInformationStructured`, so proposing a new one in the thread.

Still the same error in the console, but what i discovered now is that the error about a missing date comes from the place:
image

the `oldestTransaction` does not seem to be through normalization function 🤔 

couldn't see where do we connect the frontend with the backend the frontend calls `downloadGoCardlessTransactions` to try tracking it down for debugging 🤔

kyrias:
> sorry, @kyrias, it's still error'ing. thanks for sticking on so far muscle
> 
> The proposed regex expression does not cater for the full format of `remittanceInformationStructured`, so proposing a new one in the thread.
> 
> Still the same error in the console, but what i discovered now is that the error about a missing date comes from the place: image
> 
> the `oldestTransaction` does not seem to be through normalization function thinking
> 

I don't get how this transaction wasn't filtered out.  With it not matching the regexp `normalizeTransaction` should've returned `null`, which the `filter` call should've then filtered out.  Are you entirely sure that the server was actually running the latest version?  I can't think of any other way for it to have returned that transaction in that form.


> couldn't see where do we connect the frontend with the backend the frontend calls `downloadGoCardlessTransactions` to try tracking it down for debugging thinking

The flow is `syncGoCardlessAccount -> downloadGoCardlessTransactions -> (actual-server) app.gocardless.js /transactions handler -> getTransactionsWithBalance -> getTransactions`

src/app-gocardless/services/gocardless-service.js
```diff
@@ -438,6 +439,14 @@ export const goCardlessService = {
 
     handleGoCardlessError(response);
 
+    const bankAccount = BankFactory(institutionId);
+    response.transactions.booked = response.transactions.booked
+      .map(bankAccount.normalizeTransaction, true)
+      .filter((transaction) => transaction);
+    response.transactions.pending = response.transactions.pending
+      .map(bankAccount.normalizeTransaction, false)
+      .filter((transaction) => transaction);
```
leikoilja, line 448: haha took me a few dozen of `console.log` to realise that's where the filtering logic was failing... 😅 
`booked` boolen was never passed to `normalizeTransaction` function and instead it was the transaction index from `map` function


```suggestion
    response.transactions.booked = response.transactions.booked
      .map((transaction) => bankAccount.normalizeTransaction(transaction, true))
      .filter((transaction) => transaction);
    response.transactions.pending = response.transactions.pending
      .map((transaction) =>
        bankAccount.normalizeTransaction(transaction, false),
      )
      .filter((transaction) => transaction);
```

src/app-gocardless/services/gocardless-service.js
```diff
@@ -429,7 +430,7 @@ export const goCardlessService = {
    * @throws {ServiceError}
    * @returns {Promise<import('../gocardless.types.js').GetTransactionsResponse>}
    */
-  getTransactions: async ({ accountId, startDate, endDate }) => {
+  getTransactions: async ({ institutionId, accountId, startDate, endDate }) => {
     const response = await client.getTransactions({
       accountId,
       dateFrom: startDate,
```
leikoilja, line 436: we need to somehow remove `dataTo` parameter when calling `getTransactions`. I am not sure if we can remove it for all providers or we need to find a way to remove it for Norwegian SE alone? 🤔 
See more details in the PR comments

leikoilja (changes requested):
phew... thanks @kyrias for patience here 😅 
okay, after fixing the transaction filtering `map` logic - the sync has finally worked and the pending transactions were correctly parsed and imported, but my 'balance' was way off 🤔 
i started looking closer and realised that booked transaction were missing. which is super weird....
started digging and comparing to `raw` API calls to gocardless and finally discovered that what is messing it up with credit cards is passing optional `dateFrom` and `dateTo` params when calling `client.getTransactions`. 
Perhaps keeping `dateFrom` makes sense not to load all the transactions, but i believe we need to remove `dateTo/endDate` parameter. It's always default to `today()` but this is not a right approach (at least for Norwegian credit card). Here is why, this credit card transaction for example:
```json
            {
                "transactionId": "xxx",
                "bookingDate": "2023-08-04",
                "valueDate": "2023-08-28",
                "transactionAmount": {
                    "amount": "-47.9",
                    "currency": "SEK"
                },
                "currencyExchange": [
                    {
                        "sourceCurrency": "SEK",
                        "exchangeRate": "1.00",
                        "targetCurrency": "SEK",
                        "quotationDate": "2023-08-03T00:00:00.0000000"
                    }
                ],
                "creditorName": "COOP",
                "remittanceInformationUnstructured": "COOP",
                "additionalInformation": "Grocery Stores, Supermarkets        ",
                "proprietaryBankTransactionCode": "Köp",
                "internalTransactionId": "xxxx"
            },
```
i can see that transaction on raw API calls and Gocardless dashboard, but Actual fails to load it. Reason being is that since it is a credit card transaction it's `valueDate` is set to August 28th - which is the time when it's under "no interest" before you need to start paying an interest on that credit. However the actual purchase was made on August 4th (bookingDate). But Actual has the `dateTo` set to _today(5th August)_ when loading transaction from gocardless, which only returns transactions that have `valueDate` before _today(5th August)_ - therefore not returning the transaction mentioned above.
I have tested to remove the optional `dateTo` when we call `await client.getTransactions` and it worked as expected 💥 🔥 
@kyrias, can you help to solve it? 🤔

src/app-gocardless/services/gocardless-service.js
kyrias: D'oh, that's so silly of me.  Good catch. :D

src/app-gocardless/services/gocardless-service.js
kyrias: @fstybel @MatissJanis Is there a strong reason for having a set end date for this request?  I feel like we should be able to just stop setting one at all.

MatissJanis: Don't remember there being a strong reason. I'd be fine removing it, but in a separate PR please (so we can roll back easily if something breaks).

kyrias: Great, I'll split it out now.

kyrias: Moved to #241.

leikoilja (approved):
the changes proposed in this PR looks good to me 💥 
after https://github.com/actualbudget/actual-server/pull/241 is merged, let's rebase this branch and I will re-test E2E to verify the integration works as expected 🔥 
thank you

Inrego:
Pretty sure the issue is the same on danish branch. The issue linked in this PR is linking to a comment by me explaining the issue to begin with.
I have pasted some data there. which maybe can help you identify if it's the same?
Otherwise, what exact info do you need to confirm?

I've been in contact with GoCardless about this issue, and here's some added info I gave them:

> I checked with the transaction list on Bank Norwegians web interface. The transaction I will focus on in this post is this:
> 
> ```
> {
>   "transactionId": "274252581",
>   "bookingDate": "2023-07-03",
>   "valueDate": "2023-08-01",
>   "transactionAmount": {
>     "amount": "-397.28",
>     "currency": "DKK"
>   },
>   "currencyExchange": [
>     {
>       "sourceCurrency": "DKK",
>       "exchangeRate": "1.00",
>       "targetCurrency": "DKK",
>       "quotationDate": "2023-07-02T00:00:00.0000000"
>     }
>   ],
>   "creditorName": "OK SAEDDER               ",
>   "remittanceInformationUnstructured": "OK SAEDDER                TUREBY        DK ",
>   "additionalInformation": "Automated Fuel Dispensers           ",
>   "proprietaryBankTransactionCode": "Køb",
>   "internalTransactionId": "72fc2de1464783a431fa4b93afa93c30"
> }
> ```
> Here's what it looks like on Bank Norwegian:
> image
> 
> They have 3 different dates at play:
> 
> Dato (Date): The date for when the transaction happened (credit card used in store).
> Bogføringsdato (Booking Date): The date for when the transaction is posted.
> Rentedato (Interest Date): The date where interests will be added if the credit card balance is not paid.
> 
> Right now, it seems the mapping is made like so:
> 
> Bogføringsdato -> bookingDate (seems correct)
> Rentedato -> valueDate (seems wrong, I think it should be Dato -> valueDate?)
> 
> Please let me know what you think. Don't you agree?

And this was their reply:

> Hi, René. 
>  
> Sorry for late update. Following extensive conversations with our development team and a thorough review of our guidelines pertaining to booking dates and value dates, we have reached a conclusion. At present, we are aligned with the Berlin Group standard, and moving forward, we will maintain our adherence to this established standard:
> image
> More info in our article:[ here. ](https://nordigen.zendesk.com/hc/en-gb/articles/7899367372829-valueDate-and-bookingDate-for-transactions)
> 
> In terms of chronological order:
>  
> The booking date comes before the value date. This is because the booking date marks the point at which the transaction is officially recognized and documented by the institution, while the value date represents the effective date when the funds become available.
> The booking date is when the bank or financial institution acknowledges the transaction, and it's used for internal record-keeping and documentation.
> The value date is the date when the recipient can actually utilize the funds, and it's used to determine when interest or charges related to the transaction begin to apply.
> 
> Dato (Date) ---> Transaction date, when transaction was made
> Bogføringsdato (Booking Date) ---> booking date
> Rentedato (Interest Date) ---> value_date in credit card situation for bank norwegian this is a date when interest will be added if payment is not made.
>  
> Could you please clarify. Would you like the Interest Date parameter to be added in the data? If yes, do you want it to be added to bank pdf or in our API response? 
> If any additional questions remain, please reach out to us.

leikoilja:
@Inrego, from one transaction that you logged above - it does indeed look like a similar issue.
Can you follow [this message](https://github.com/actualbudget/actual/issues/1392#issuecomment-1658921138) and log `"Available account properties for new institution integration"` in the same issue https://github.com/actualbudget/actual/issues/1392? then we should be able to adopt the solution proposed in this PR for your danish bank branch so you could test

kyrias:
Ah, I had not seen their page about these dates before, that's interesting.  I still feel like their descriptions do not really quite correspond to the Berlin Group's NextGenPSD2 definitions, but oh well.  It also means that I have some banks that I want to report to them as not being implemented according to their own documentation.

> Dato (Date) ---> Transaction date, when transaction was made
> Bogføringsdato (Booking Date) ---> booking date
> Rentedato (Interest Date) ---> value_date in credit card situation for bank norwegian this is a date when interest will be added if payment is not made.

@Inrego Their table contains `bookingDate <= valueDate <= Today` which means that putting the interest date into the `valueDate` field is violating their documentation since the interest date is many days into the future and therefore definitely not `<= Today`.  So please ask them whether the table or their integration with Bank Norwegian is incorrect.

Their table also describes `bookingDate` as "The date when the payment is initiated" which I would argue corresponds more to "Date" than to "Booking Date".

@MatissJanis With their documented definition in mind ("The date when the payment is initiated" and "bookingDate <= valueDate <= Today"), I again think it would make sense for Actual to prefer `bookingDate` over `valueDate`.  If you agree then I'll open a PR performing this change.  Additionally, yesterday I suggested that maybe the choice of whether to use the `bookingDate` or the `valueDate` actually belongs in the server along with the specific bank integration rather than in the Actual app.  This would normalize the data more and means that Actual doesn't need to know about the specific semantics of fields returned from GoCardless.  What do you think about this suggestion?

Inrego:
I've written them back:

> Thank you for clarifying. However, your own documentation contradicts your conclusion that the data returned from your api response is correct.
> In the sample provided to you on July 5th, your api returns "valueDate": "2023-08-01"
> According to the definition you just sent me: bookingDate <= valueDate <= Today
> Which means neither of the dates should be at a point in the future. But your api did indeed return a date from the future.
> 
> On the link to your documentation for valueDate and bookingDate, it is mentioned that valueDate can be a date in the future for pending transactions. But all transactions have a valueDate set to the first of the month coming after the transaction. The transactions aren't pending for a whole month.

Let's see what they say. Their response times are quite slow on the matter. I doubt they will change anything on their end.

kyrias:
I've changed the bank interface to provide an array of institution IDs and added all of the Bank Norwegian institutions.

leikoilja:
awesome, now that both https://github.com/actualbudget/actual-server/pull/241 and https://github.com/actualbudget/actual-server/pull/239 are merged, @kyrias, please let me know when you rebase this PR - i will go ahead to test the Norwegian bank implementation to verify it's implementation 🚀 
thanks 💪

MatissJanis:
> @MatissJanis With their documented definition in mind ("The date when the payment is initiated" and "bookingDate <= valueDate <= Today"), I again think it would make sense for Actual to prefer bookingDate over valueDate. If you agree then I'll open a PR performing this change. Additionally, yesterday I suggested that maybe the choice of whether to use the bookingDate or the valueDate actually belongs in the server along with the specific bank integration rather than in the Actual app. This would normalize the data more and means that Actual doesn't need to know about the specific semantics of fields returned from GoCardless. What do you think about this suggestion?


1. changing from `valueDate` to `bookingDate` - I'm not sure why we chose `valueDate` before, but I'm fine with this changing to `bookingDate`
2. moving the normalization from UI to server: makes sense

Inrego:
GoCardLess has replied already:

> I have to apologize about some inconsistencies in our documentation.
> That schema is for cash account transactions only.
> We will add this section, about credit cards ASAP, so in the future there won't be any misunderstanding.
> In this situation that you mentioned :
> value_date == interest date, that is the date until which you can clear the funds and no interest will be added. In most credit card cases the bank gives you time till the 1st date of the next month to do so, but there are edge cases for bigger purchases and in Bank Norwegian case period can be up to 45days.
> I hope that this will clear things out for you.

I've replied again, asking to confirm that it is not possible to get the date of the transaction from their api.

leikoilja (approved):
Boom, I can confirm that the integration provider Norwegian SE works perfectly as expected now 💥 🔥 
Also the code changes look good to me, thanks for adding the support for all Norwegian Bank branches.
Muchas gracias, @kyrias 💪

Inrego:
How do I test this? Is it on a Docker tag?

leikoilja:
@Inrego, git clone both repositories separately.This and actual-server.
For this 'actual' repository checkout this branch (the one this PR is open for). [some quick guide](https://devopscube.com/checkout-git-pull-request/)
In each directories run yarn install and yarn start. That will spawn you a local Actual instance. In that add a new linked account to your bank and follow the instructions.

Inrego:
I'm sorry, but with a baby and a toddler in the home, I'm afraid I won't have the time to go through the steps of setting up a test environment from this PR. I'll just have to wait until it's published on a Docker tag I can simply pull.

PR #239, opened 2023-08-05
Add GoCardless bank integration for American Express AESUDEF1

kyrias:
The first commit has the same transaction normalization commit as #237.


MatissJanis (approved):
Good idea of adding `normalizeTransaction`. That will be useful.

As for the specific american express bank integration: rubber stamp approval there as I don't use this bank myself.

PR #242, opened 2023-08-09
Add FINTRO_BE_GEBABEBB GoCardless integration (Include additionalInformation in notes)

CharlieMK:
Implements https://github.com/actualbudget/actual/issues/1466
Replaces https://github.com/actualbudget/actual/pull/1490
Uses @kyrias transaction normalization from https://github.com/actualbudget/actual-server/pull/237

Note that due to https://github.com/actualbudget/actual/pull/745, in case of **pending** transactions 'atmPosName' will be used for both the payeename and the notes field. This is because pending transactions do not have a payee name yet, but they do have an atmPosName in the additionalInformation field that I'm using. I would argue that we should not blindly put remittanceinformation in the payee name field (except perhaps for some banks), but I'm not removing that in this PR.


MatissJanis:
👋 I'll review this after https://github.com/actualbudget/actual-server/pull/237 is merged.

CharlieMK:
Rebased because 237 was merged.

MatissJanis (approved):

PR #243, opened 2023-08-09
Decide on transaction date during GoCardless transaction normalization

kyrias:
Previously the Actual backend would default to using `valueDate` and
falling back to `bookingDate`.  According to the GoCardless
documentation[^0] we should prefer `bookingDate` if it's set, but the
different bank integrations vary greatly in how they use the date fields
and so for consistency we should instead let the Actual bank
integrations decide which is the correct date for any given transaction.

For any transaction without an appropriate date to return the
`normalizeTransaction` function should return `null` so that it's
filtered out until the bank provides the full transaction details.

The existing bank integrations for ING_PL_INGBPLPW and
MBANK_RETAIL_BREXPLPW were changed to retain the existing precedence
order, though I'm not sure which is the correct one for these banks.

  [^0]: https://nordigen.zendesk.com/hc/en-gb/articles/7899367372829-valueDate-and-bookingDate-for-transactions

### DIFF
--- a/src/app-gocardless/banks/integration-bank.js
+++ b/src/app-gocardless/banks/integration-bank.js
@@ -1,3 +1,4 @@
+import * as d from 'date-fns';
 import {
   sortByBookingDateOrValueDate,
   amountToInteger,
@@ -37,7 +38,21 @@ export default {
   },
 
   normalizeTransaction(transaction, _booked) {
-    return transaction;
+    const date =
+      transaction.bookingDate ||
+      transaction.bookingDateTime ||
+      transaction.valueDate ||
+      transaction.valueDateTime;
+    // If we couldn't find a valid date field we filter out this transaction
+    // and hope that we will import it again once the bank has processed the
+    // transaction further.
+    if (!date) {
+      return null;
+    }
+    return {
+      ...transaction,
+      date: d.format(d.parseISO(date), 'yyyy-MM-dd'),
+    };
   },
 
   sortTransactions(transactions = []) {
--- a/src/app-gocardless/banks/sandboxfinance-sfin0000.js
+++ b/src/app-gocardless/banks/sandboxfinance-sfin0000.js
@@ -36,12 +36,7 @@ export default {
   },
 
   sortTransactions(transactions = []) {
-    return transactions.sort((a, b) => {
-      const [aTime, aSeq] = a.transactionId.split('-');
-      const [bTime, bSeq] = b.transactionId.split('-');
-
-      return Number(bTime) - Number(aTime) || Number(bSeq) - Number(aSeq);
-    });
+    return sortByBookingDateOrValueDate(transactions);
   },
 
   /**
--- a/src/app-gocardless/banks/sandboxfinance-sfin0000.js
+++ b/src/app-gocardless/banks/sandboxfinance-sfin0000.js
@@ -1,4 +1,8 @@
-import { printIban, amountToInteger } from '../utils.js';
+import {
+  printIban,
+  amountToInteger,
+  sortByBookingDateOrValueDate,
+} from '../utils.js';
 
 /** @type {import('./bank.interface.js').IBank} */
 export default {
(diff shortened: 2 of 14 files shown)


src/app-gocardless/banks/integration-bank.js
```diff
@@ -37,7 +37,14 @@ export default {
...
   normalizeTransaction(transaction, _booked) {
-    return transaction;
+    return {
+      ...transaction,
+      date:
+        transaction.bookingDate ||
+        transaction.bookingDateTime ||
+        transaction.valueDate ||
+        transaction.valueDateTime,
+    };
```
MatissJanis, line 47: **💬 suggestion:** ‏I think the ordering used here will cause a regression. UI code: https://github.com/actualbudget/actual/pull/1499/files

```suggestion
      date:
        transaction.valueDate ||
        transaction.valueDateTime ||
        transaction.bookingDate ||
        transaction.bookingDateTime,
    };
```

src/app-gocardless/banks/sandboxfinance-sfin0000.js
```diff
@@ -16,17 +20,23 @@ export default {
...
+   * differently from what's describen in their documentation and so it's
+   * sometimes necessary to use `valueDate` instead.
+   *
+   *   [0]: https://nordigen.zendesk.com/hc/en-gb/articles/7899367372829-valueDate-and-bookingDate-for-transactions
+   */
   normalizeTransaction(transaction, _booked) {
-    return transaction;
+    return {
+      ...transaction,
+      date: transaction.bookingDate || transaction.valueDate,
```
MatissJanis, line 34: **💬 suggestion:** ‏here too

```suggestion
      date: transaction.valueDate || transaction.bookingDate,
```

MatissJanis (commented):
Overall LGTM! Just two changes in functionality I think should be fixed.

src/app-gocardless/banks/integration-bank.js
kyrias: It's a change in behavior, but I don't see how it would be a regression since we're switching to the expected order as documented by GoCardless and which you seemed okay with here: https://github.com/actualbudget/actual-server/pull/237#issuecomment-1670822085

I'd also be fine with using the minimum as you suggested, but I figured that using the order that I'd expect according to their documentation would be more semantically correct.

MatissJanis: Apologies; poor choice of words of my behalf. Yes, this isn't a regression, but a change in functionality.

Conceptually I agree with the order you are proposed, but I'm worried about changing the functionality unless there is a good reason for the change as it introduces a certain factor of risk? If there is no specific reason to change the logic (bug/issue) - I'd err on the side of keeping the order as-is until we find a good reason for the change. Thus having it more stable.

WDYT?

kyrias: The one risk is that some banks that violate their documentation will have transactions returned with a later date than previously (by a day or two), but on the other hand banks that _don't_ violate their documentation will start returning the correct data which they until now haven't.

So while I don't necessarily have a problem with leaving it as-is it means that we're choosing to give priority to users of banks where the API returns 'incorrect' data over users of banks where the API returns 'correct' data.

Which to me seems strange when the only thing that the affected users will experience is that their imported transactions, which they haven't manually previously created a transaction that the imported one will be merged into, will be off by a day or two.  I would argue that this would also be a good hint to them that they should file a request for an integration with their bank to be added so that its incorrect behavior can be explicitly normalized rather than just accidentally working at the expense of banks that do the 'correct' thing.

---

Ultimately I'm planning on adding integrations for all banks that I care about so it won't affect me directly, and so I don't really care too much about which way we go other than from a theoretical correctness perspective.

kyrias: (I say that, but as written this is actually wrong I just noticed.  The `DateTime` variants need to be parsed since we expect only dates.)

MatissJanis: Hmm.. ok, you convinced me. Worst case scenario: this breaks in which case we can quickly patch it.

But can we then make all the banks follow this order? For example: ing-pl-ingbplpw is following the old ordering. Whichever way we choose - we should make it consistent.

kyrias: Since the previous integrations (ing-pl-ingbplpw and mbank-retail-brexplpw) were added as part of the initial Nordigen support I assumed that the old order was the correct one for those banks, and that they'd have either added transaction normalization or just defaulted to the other order if it wasn't the correct order for them.  I don't have any example transactions from those banks to go off of though so I can't be sure, and so I can change it to the new order if you'd like.

MatissJanis: I would say if we're changing this logic in one place, then lets change it in all. Lets keep the code consistent. If after merging someone reports this as being the wrong call - we can always swap back. But I doubt it will be this case.

kyrias: Done.

MatissJanis (approved):
Thanks!

PR #244, opened 2023-08-10
Add DANSKEBANK_DABANO22 integration

LudvigHz:
This PR adds an integration for `DANSKEBANK_DABANO22`.

The issue here is mostly similar to that fixed in #237 (`valueDate` is handled similarly), except from the issues with pending transactions (I _think_).

I'm going to take a wild guess here and say that the other branches of Danske Bank private (`DANSKEBANK_DABAFIHH`, `DANSKEBANK_DABADKKK`, `DANSKEBANK_DABASESX`, `DANSKEBANK_DABAGB2B`) _most likely_ exhibit the same behavior? But since I can't validate this I've just enabled this for the NO branch currently. Would love some feedback on this.

Another issue here is that I've only validated this behavior for the credit account. I have a checking account but don't really use it so I have no data on what that looks like, although I would expect it to not have the same issues with `valueDate`. 

Let me know if I should fall back to the default there or not. The only issue is that I would have to refactor a bit to get information about the account in order to differentiate the transactions between the account types. If it's easier to just keep `bookingDate` for all accounts I feel like that would be fine. But I can of course change that.

Regarding pending transactions, I have some in my bank right now I believe (they are marked as `reserved` in by banking app at least, but they still show up under booked transactions from GoCardless. So my guess here as well is that they just put everything in `booked` regardless of their actual status :facepalm:.

### DIFF
--- a/src/app-gocardless/banks/danskebank-dabno22.js
+++ b/src/app-gocardless/banks/danskebank-dabno22.js
@@ -1,0 +1,60 @@
+import {
+  printIban,
+  amountToInteger,
+  sortByBookingDateOrValueDate,
+} from '../utils.js';
+
+/** @type {import('./bank.interface.js').IBank} */
+export default {
+  institutionIds: ['DANSKEBANK_DABANO22'],
+
+  normalizeAccount(account) {
+    return {
+      account_id: account.id,
+      institution: account.institution,
+      mask: account.iban.slice(-4),
+      iban: account.iban,
+      name: [account.name, printIban(account)].join(' '),
+      official_name: account.name,
+      type: 'checking',
+    };
+  },
+
+  normalizeTransaction(transaction, _booked) {
+    /**
+     * Danske Bank appends the EndToEndID: NOTPROVIDED to
+     * remittanceInformationUnstructured, cluttering the data.
+     *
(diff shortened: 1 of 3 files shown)


LudvigHz:
I just noticed #243, which fixes the date inconsistency :)

Although I'd still like to keep this to clean up the nasty `EndToEndId: NOTPROVIDED` stuff.

kyrias (approved):
This looks reasonable to me.

If this gets merged before my date selection PR then I'll update this integration there.

src/app-gocardless/banks/danskebank-dabno22.js
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * We clean thais up by removing any instances of this string from all transactions.
+     *
+     */
+    transaction.remittanceInformationUnstructured =
+      transaction.remittanceInformationUnstructured.replace(
+        '\nEndToEndID: NOTPROVIDED',
+        '',
+      );
+    return transaction;
```
MatissJanis, line 42: **💬 suggestion:** ‏
```suggestion
    return {
      ...transaction,
      date: transaction.bookingDate,
    };
```

and then you can also remove line 28

MatissJanis (approved):
Thanks!

Issue #246, opened 2023-08-15
Default data dir on Windows


UnexomWid:
Hello!

I'm running the server locally on Windows, and I've encountered a small issue regarding the default data dir location.

When not using the ACTUAL_CONFIG_PATH env var, I would expect the server to load the config from the project root. What actually happens is this:

- when loading the config, the default data dir is set to `/data` if it exists
- on Windows, if you run the server locally on `D:/server/path/here`, this default dir `/data` is actually `D:/data`
- in my case, `D:/data` happens to exist but contains files completely irrelevant to Actual
- the server tries to open `D:/data/config.json`, which doesn't exist
- the config isn't loaded, and therefore the server creates `D:/data/server-files` and `D:/data/user-files`

I expected the server to instead use the config file in the project root.

1. Is this intended?
2. If not, I can create a PR and change the behavior by either:
   - falling back on the project root if `config.json` doesn't exist in `/data`, or
   - directly setting the default data dir to the project root only on Windows (since I can't think of any use cases for `/data` on Windows, but I could be wrong), or
   - other suggestions

MatissJanis:
👋 Very good observation!

IMO the best solution would be to pick up the `config.json` file from project root. And if it's not there - fall back to `/data` path. I find it actually a bit weird that project root isn't the default option for it already...

Anyways: feel free to submit a PR for this. I'd be happy to approve and merge :)

MatissJanis:
Closing as DONE

PR #249, opened 2023-08-16
Fix config file not being loaded from the project root by default

UnexomWid:
Solves #246 by loading the config from the project root if it exists there.

### DIFF
--- a/src/load-config.js
+++ b/src/load-config.js
@@ -10,7 +10,10 @@ const projectRoot = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
 debug(`project root: '${projectRoot}'`);
 export const sqlDir = path.join(projectRoot, 'src', 'sql');
 
-let defaultDataDir = fs.existsSync(path.join(projectRoot, 'config.json')) ? projectRoot : '/data';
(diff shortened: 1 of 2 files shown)


src/load-config.js
```diff
@@ -10,7 +10,10 @@ const projectRoot = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
 debug(`project root: '${projectRoot}'`);
 export const sqlDir = path.join(projectRoot, 'src', 'sql');
 
-let defaultDataDir = fs.existsSync('/data') ? '/data' : projectRoot;
+let defaultDataDir = fs.existsSync(path.join(projectRoot, 'config.json'))
+  ? projectRoot
+  : '/data';
```
MatissJanis, line 15: Ok, this won't really work. So folks can define a custom `data` (database) location. For example: dropbox. And we should always respect the users choice in this.

However `config.json` is different. For this specific file we could implement the logic you are proposing here. But only for `config.json` file. Not for `/data`.

Does that make sense?

UnexomWid: I see, thanks for bringing that up. I'll fix this shortly.

PR #257, opened 2023-08-30
Add GoCardless bank integration for Spar Nord SPARNORD_SPNODK22

Aarup:
This adds support for importing transactions from Spar Nord. The only extra information regarding a transaction if contained in AdditionalInformation, so this is mapped to remittanceInformationUnstructured


MatissJanis (approved):
Thanks! I'll merge this after merge freeze is lifted

PR #265, opened 2023-09-08
Make edge builds fetch master builds from correct repository

kyrias:
If we don't verify that the repository ID is what we expected then we will pull artifacts from PRs where the source branch is `master`.


MatissJanis:
Thanks!

Could someone who has docker experience please verify this patch? I'm not very experienced here.

shall0pass (approved):
Thanks for this!  I've confirmed that the proposed change only matches to artifacts where the workflow run is initiated from the actualbudget repo's master branch.

PR #267, opened 2023-09-10
:sparkles:  ability to add migrations

### LINKED ISSUE #236: [Tech Debt] Ability to run DB migrations
It would be great if `actual-server` exposed a way to run DB migrations on the **server database**.

https://github.com/actualbudget/actual-server/pull/219 has a migration implementation, but only for the `accounts` and `secrets` table (missing `messages` table). Would be great to expand it to all **server tables**.

MatissJanis:
Ability to create DB migrations in `/migrations` folder.

Also small code cleanups for the existing "migrations" that we had in place - to move to the new approach.

Closes #236


src/app-gocardless/services/gocardless-service.js
```diff
@@ -17,17 +17,12 @@ import jwt from 'jws';
...
-});
-secretsService.onUpdate(SecretName.nordigen_secretKey, (newSecret) => {
-  goCardlessClient.secretKey = newSecret;
-});
+
+const getGocardlessClient = () =>
+  new GoCardlessClient({
+    secretId: secretsService.get(SecretName.nordigen_secretId),
+    secretKey: secretsService.get(SecretName.nordigen_secretKey),
+  });
```
MatissJanis, line 25: Small amount of refactors here. Just cleaning up the code by reducing the complexity.

PR #274, opened 2023-11-04
Add "docker.io/" registry prefix to docker compose image

Valloric:
Required by Podman Compose. Docker Compose should not care either way. The registry prefix is explicitly supported in the [compose file spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#image)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   actual_server:
-    image: actualbudget/actual-server:latest
+    image: docker.io/actualbudget/actual-server:latest
     ports:
       # This line makes Actual available at port 5006 of the device you run the server on,
       # i.e. http://localhost:5006. You can change the first number to change the port, if you want.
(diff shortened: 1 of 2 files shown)


joel-jeremy:
@MatissJanis do you think we should point the docker compose file to the GitHub container registry or should we use docker hub instead?

joel-jeremy:
@Valloric actual server images are also hosted on docker hub. Can we point the compose file there instead?

Valloric:
I changed the prefix to point to docker hub and added the release notes. Should be good to go now.

Some friendly advice from someone who spent 15+ years maintaining various OSS projects: the [release notes process](https://actualbudget.org/docs/contributing/#writing-good-release-notes) adds needless overhead and will cut the rate of PRs you get by a factor of 10. This process is largely automatable at the time of cutting a release. The non-automatable parts (categorization of bugfix/feature/whatever) is not worth the overhead.

Do with that advice as you wish. :)

joel-jeremy (approved):

joel-jeremy:
Thank you! We'll keep that advice in mind. :)

PR #275, opened 2023-11-08
:bug: (docker) add missing migrations folder

MatissJanis:


shall0pass (approved):

shaankhosla:
Might be missing something, but doesn't this change also need to be added to the four Dockerfiles in the docker folder?

MatissJanis:
Good catch @shaankhosla ! Updated those too

(as you can clearly see - docker isn't exactly my strong suit :) )

youngcw (approved):

PR #278, opened 2023-11-15
:bug: (gocardless) fix banksync - create singleton client

MatissJanis:
https://github.com/actualbudget/actual-server/pull/267 introduced a small regression: the client is re-created every time it is used. Obviously this causes some problems as the access tokens get reset very often..

Instead I am now using a singleton client that is mapped by the secrets. Meaning: when secrets change we will generate a new client.

### DIFF
--- a/src/app-gocardless/services/gocardless-service.js
+++ b/src/app-gocardless/services/gocardless-service.js
@@ -18,11 +18,22 @@ import { SecretName, secretsService } from '../../services/secrets-service.js';
 
 const GoCardlessClient = nordigenNode.default;
 
-const getGocardlessClient = () =>
-  new GoCardlessClient({
+const clients = new Map();
+
+const getGocardlessClient = () => {
+  const secrets = {
     secretId: secretsService.get(SecretName.nordigen_secretId),
     secretKey: secretsService.get(SecretName.nordigen_secretKey),
-  });
+  };
+
+  const hash = JSON.stringify(secrets);
+
+  if (!clients.has(hash)) {
+    clients.set(hash, new GoCardlessClient(secrets));
+  }
+
+  return clients.get(hash);
+};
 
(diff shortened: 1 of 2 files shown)


src/app-gocardless/services/gocardless-service.js
```diff
@@ -18,11 +18,22 @@ import { SecretName, secretsService } from '../../services/secrets-service.js';
 
 const GoCardlessClient = nordigenNode.default;
 
-const getGocardlessClient = () =>
-  new GoCardlessClient({
+const clients = new Map();
```
toebbel, line 21: Can different budgets have different Nordigen secrets? I guess the `Map` makes sense here 👍

src/app-gocardless/services/gocardless-service.js
```diff
@@ -18,11 +18,22 @@ import { SecretName, secretsService } from '../../services/secrets-service.js';
...
     secretId: secretsService.get(SecretName.nordigen_secretId),
     secretKey: secretsService.get(SecretName.nordigen_secretKey),
-  });
+  };
+
+  const hash = JSON.stringify(secrets);
+
+  if (!clients.has(hash)) {
+    clients.set(hash, new GoCardlessClient(secrets));
+  }
```
toebbel, line 33: 💯  nice!

toebbel (approved):
awesome, thanks for the quick support 🙏 
I tested this branch on my fly instance and the bank sync is fixed :)

youngcw (approved):

src/app-gocardless/services/gocardless-service.js
MatissJanis: Different budgets can have different secrets, but also the same budget might change its keys. So in that case too we also want to generate a new client.

PR #283, opened 2023-11-27
Add workaround for non-unique transactionIds for Belfius bank

nudded:
Belfius bank has non-unique transaction ids causing data to not be synced. 

This is documented in the gocardless api and there is an internalTransactionId field available which gocardless makes sure is unique.


MatissJanis (approved):
Thanks

PR #285, opened 2023-12-01
Changed fintro specific bank integration to work with BNP and Hello Bank

feliciaan:
Change the gocardless integration that works for Fintro (subbrand) to also work with the main institution (BNP) and another brand of the bank (Hello Bank). Also rename of the file to the main brand.


MatissJanis (approved):
👍

PR #289, opened 2023-12-04
Store migrations statestore in datadir

### LINKED ISSUE #288: [Bug]: 23.12.0 rootless migrate permission error

### What happened?

After Upgrading to 23.12.0 the following permission error appears when running rootless (as before).

I am running actual in k3s with this pod securityContext:
```
  fsGroup: 64087
  runAsGroup: 64087
  runAsNonRoot: true
  runAsUser: 64087
  seccompProfile:
    type: RuntimeDefault
```
and container securityContext:
```
  allowPrivilegeEscalation: false
  capabilities:
    drop:
      - ALL
  runAsGroup: 64087
  runAsNonRoot: true
  runAsUser: 64087
```

I think the problem is the app tries to create a `.migrate` file in `/app` not in `/data`.

```
Checking if there are any migrations to run for direction "up"...
file:///app/src/migrations.js:21
            throw err;
            ^

[Error: EACCES: permission denied, open '.migrate'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'open',
  path: '.migrate'
}

Node.js v18.19.0

```

Other

Chrome

### Operating System

Linux

bjw-s:
This PR makes it so the `.migrations` file gets stored in the `dataDir` instead of the application root. This should fix #288 and https://github.com/actualbudget/actual/issues/2011.

### DIFF
--- a/src/config-types.ts
+++ b/src/config-types.ts
@@ -2,6 +2,7 @@ import { ServerOptions } from 'https';
 
 export interface Config {
   mode: 'test' | 'development';
+  dataDir: string;
   port: number;
   hostname: string;
   serverFiles: string;
--- a/src/load-config.js
+++ b/src/load-config.js
@@ -44,7 +44,7 @@ if (process.env.ACTUAL_CONFIG_PATH) {
   userConfig = parseJSON(configFile, true);
 }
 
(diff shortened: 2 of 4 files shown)


MatissJanis (approved):
Thanks!

PR #293, opened 2023-12-15
:recycle: (gocardless) renamed nordigen_* secrets to gocardless_*

MatissJanis:

### DIFF
--- a/migrations/1702667624000-rename-nordigen-secrets.js
+++ b/migrations/1702667624000-rename-nordigen-secrets.js
@@ -1,0 +1,19 @@
+import getAccountDb from '../src/account-db.js';
+
+export const up = async function () {
+  await getAccountDb().exec(
+    `UPDATE secrets SET name = "gocardless_secretId" WHERE name = "nordigen_secretId"`,
+  );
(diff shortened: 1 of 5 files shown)


twk3:
@MatissJanis is there a matching MR for updating the `secret-set` in desktop-client? It's still pointing at `nordigen` there. Currently with this change you can no longer init gocardless fresh.

Might be easiest to leave both sets of constants here in  `SecretName`, both pointing to the new names. Then following up with a final MR to remove the extra constants after the update in desktop-client is released.

MatissJanis:
Good catch @twk3 !

We can update both of these variables in-sync. Here's the web PR: https://github.com/actualbudget/actual/pull/2181

youngcw (approved):

twk3 (approved):
Looks good! Ship it!. 

I tested with the linked actual-web PR, and confirmed the migrations worked for secrets setup before the change, and that fresh setups also work after the change.

Also did a quick sanity that gocardless bank setup was still working using the sandbox institution https://developer.gocardless.com/bank-account-data/sandbox

PR #296, opened 2024-01-07
SimpleFin

zachwhelchel:
SimpleFin can now be configured in the UI and sync transactions, somewhat adjacent to this [issue](https://github.com/actualbudget/actual/issues/898).

There is also an app PR for these changes: https://github.com/actualbudget/actual/pull/2188

General discussion should be held on the app PR linked above.


src/app-simplefin/app-simplefin.js
```diff
@@ -0,0 +1,236 @@
...
+
+const app = express();
+export { app as handlers };
+app.use(express.json());
+
+app.post('/status', async (req, res) => {
+  let configured = false;
+
+  let token = secretsService.get(SecretName.simplefin_token);
+  if (token !== null && token !== undefined && token !== 'Forbidden') {
```
joel-jeremy, line 14: These checks can be simplified to just `token != null` which check for both null and undefined.

src/app-simplefin/app-simplefin.js
```diff
@@ -0,0 +1,236 @@
...
+        referenceDate: date,
+      },
+    ];
+    response.iban = 'thisismadeup';
+    response.institutionId = 'thisismadeup';
+    response.startingBalance = balance; // could be named differently in this use case.
+
+    let allTransactions = [];
+
+    for (let i = 0; i < account.transactions.length; i++) {
```
joel-jeremy, line 100: We can use for..in here :)

src/app-simplefin/app-simplefin.js
```diff
@@ -0,0 +1,236 @@
...
+        balanceType: 'expected',
+        referenceDate: date,
+      },
+      {
+        balanceAmount: { amount: account.balance, currency: account.currency },
+        balanceType: 'interimAvailable',
+        referenceDate: date,
+      },
+    ];
+    response.iban = 'thisismadeup';
```
joel-jeremy, line 94: Is this intentional?

zachwhelchel:
@joel-jeremy @youngcw @MatissJanis this PR is ready as well now.

PR #300, opened 2024-01-14
non-root users for Dockerfiles

hkiang01:
---
category: Maintenance
authors: hkiang01
---

Non-root users for Dockerfiles

As a regular Node application, the default user doesn't need to be root. For why this is important, see [Why non-root containers are important for security](https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security)

### DIFF
--- a/upcoming-release-notes/300.md
+++ b/upcoming-release-notes/300.md
@@ -3,4 +3,4 @@ category: Maintenance
 authors: [hkiang01]
 ---
 
-Non-root users for Dockerfiles
+Optional non-root user for Docker
--- a/upcoming-release-notes/300.md
+++ b/upcoming-release-notes/300.md
@@ -3,4 +3,4 @@ category: Maintenance
 authors: [hkiang01]
 ---
 
-Non-root users for stable Dockerfiles
+Non-root users for Dockerfiles
--- a/upcoming-release-notes/300.md
+++ b/upcoming-release-notes/300.md
@@ -1,0 +1,6 @@
+---
+category: Maintenance
+authors: [hkiang01]
+---
+
+Non-root users for stable Dockerfiles
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN yarn workspaces focus --all --production
 FROM node:18-bullseye-slim as prod
 RUN apt-get update && apt-get install tini && apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
-ARG USERNAME=appuser
+ARG USERNAME=actual
 ARG USER_UID=1001
 ARG USER_GID=$USER_UID
 RUN groupadd --gid $USER_GID $USERNAME \
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,11 +7,20 @@ RUN yarn workspaces focus --all --production
 
 FROM node:18-bullseye-slim as prod
 RUN apt-get update && apt-get install tini && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+
+ARG USERNAME=appuser
+ARG USER_UID=1001
+ARG USER_GID=$USER_UID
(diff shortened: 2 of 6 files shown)


hkiang01:
I tested this locally by altering docker-compose.yml to point to each edited Dockerfile. I was able to successfully build and start each corresponding built container to serve localhost:5006

hkiang01:
hmmm got an error while syncing

```
Rejection: SqliteError: attempt to write a readonly database
    at WrappedDatabase.mutate (file:///app/src/db.js:39:21)
    at file:///app/src/sync-simple.js:31:23
    at sqliteTransaction (/app/node_modules/better-sqlite3/lib/methods/transaction.js:65:24)
    at WrappedDatabase.transaction (file:///app/src/db.js:47:35)
    at addMessages (file:///app/src/sync-simple.js:26:6)
    at Module.sync (file:///app/src/sync-simple.js:81:14)
    at file:///app/src/app-sync.js:111:42
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at next (/app/node_modules/express/lib/router/route.js:144:13)
    at Route.dispatch (/app/node_modules/express/lib/router/route.js:114:3) {
  code: 'SQLITE_READONLY'
  ```

**Review by twk3 — CHANGES_REQUESTED** [review]
@hkiang01 the app code should remain root, so the runtime user can't manipulate the code. Runtime users should just be able to influence the config and data dir.

**Comment by twk3** [comment]
> I recommend bumping major version as the non-root user wouldn't have access to read/write files owned by root. This can be fixed by running `chown -R 1001:1001 /data` in the container

This is a good point, but I think that's a bit of a blocker for us for now. We also have to consider users already running on pikapods and fly.io

@hkiang01 I think as an iteration, adding the user, but not setting it as the run user for entry. Chowning the data directory contents on entry to the new user, and then running the app as the new user will set us up to switchover in a major release.

**Comment by hkiang01** [comment]
Requested changes made. I expect a retry of the build jobs will succeed. Please retry the jobs and re-review

**Comment by hkiang01** [comment]
failing builds are due to #287 which is out of scope of this PR

**Comment by hkiang01** [comment]
confirmed non-breaking change with manual steps:

1. spin up container with default user (root)
2. navigate to localhost:5006, perform some actions, sync
3. re-create container with user `appuser`
4. repeat step 2

File: upcoming-release-notes/300.md [hunk 1]
```diff
@@ -0,0 +1,6 @@
+---
+category: Maintenance
+authors: [hkiang01]
+---
+
+Non-root users for Dockerfiles
```
twk3, line 6:
```suggestion
Optional Non-root user for Docker
```

twk3 (changes requested):
@hkiang01 this looks like a good first step, and I think we can move forward with this change after a few minor tweaks for users who want to set the user to 1001 when running from their very first run. (Like in your helm chart). But the swap from root to app user does not currently work without manual chowning of the data. (database files will be readonly by the appuser)

A followup (doesn't have to block this PR). Would be to move our current RUN command to a startup script, then earlier in that script check if the running  uid is 0, if so, chown all data (at runtime this time) to our non-root user, and then exec our tini command as that user. Else we just exec the tini command like normal.

Bitnami has some examples of using chroot to change the user at runtime if running as root by just chrooting to `/` again but with userspace. https://github.com/bitnami/containers/blob/8a67664e7a937c6aa603a9b0477561b9af7abb0e/bitnami/postgresql/16/debian-11/prebuildfs/opt/bitnami/scripts/libos.sh#L653

This would allow us to get everyone running as non-root, and in a later major change we could just drop the extra work for the USER line in the dockerfile.

twk3 (approved):
Thanks @hkiang01 

This is a good first step towards moving the container to always run non-root.

Issue #287, opened 2023-12-02
[Bug]: CI - out of memory errors


MatissJanis:
### Verified issue does not already exist?

- [X] I have searched and found no existing issue

### What happened?

The edge docker build CI jobs sometimes randomly fail. This is especially annoying for the scheduled `edge` releases we have every day. Sometimes consecutive days fail running, so we don't get as frequent releases. I haven't been able to figure out why.. 

Example: https://github.com/actualbudget/actual-server/actions/runs/7066898754/job/19239533606

```
#51 [linux/arm/v7 base  6/12] RUN yarn workspaces focus --all --production
#51 11.21 ➤ YN0000: ┌ Resolution step
#51 12.19 ➤ YN0085: │ - @ampproject/remapping@npm:2.2.0, @babel/code-frame@npm:7.18.6, @babel/code-frame@npm:7.22.13, @babel/compat-data@npm:7.20.14, @babel/core@npm:7.20.12, and 384 more.
#51 12.28 ➤ YN0000: └ Completed in 1s 67ms
#51 12.30 ➤ YN0000: ┌ Fetch step
#51 50.61 ➤ YN0001: │ Error: @actual-app/web@npm:23.11.0: Couldn't allocate enough memory
#51 50.61     at Ji.allocateBuffer (/app/.yarn/releases/yarn-4.0.1.cjs:149:204491)
#51 50.61     at Ji.allocateUnattachedSource (/app/.yarn/releases/yarn-4.0.1.cjs:149:204733)
#51 50.61     at new Ji (/app/.yarn/releases/yarn-4.0.1.cjs:149:193911)
#51 50.61     at Object.sut (/app/.yarn/releases/yarn-4.0.1.cjs:192:212275)
#51 50.61     at async dl.fetchFromNetwork (/app/.yarn/releases/yarn-4.0.1.cjs:678:3118)
#51 50.61     at async v (/app/.yarn/releases/yarn-4.0.1.cjs:199:1576)
#51 50.61     at async x (/app/.yarn/releases/yarn-4.0.1.cjs:199:1916)
#51 50.61     at async E (/app/.yarn/releases/yarn-4.0.1.cjs:199:3257)
#51 50.61     at async Lr.fetchPackageFromCache (/app/.yarn/releases/yarn-4.0.1.cjs:199:3378)
#51 50.61     at async dl.fetch (/app/.yarn/releases/yarn-4.0.1.cjs:678:2469)
#51 ...

[...]

PR #301, opened 2024-01-17
[WIP]Set the docker buildx max parallel to 2

twk3:
- helps keep our builds within the machine memory

Potential fix for https://github.com/actualbudget/actual-server/issues/287


twk3:
The max-parallel here didn't help. The yarn changes looked like they may have. Pulling them out into https://github.com/actualbudget/actual-server/pull/302 for further investigation.

PR #304, opened 2024-01-18
update os packages to resolve CVEs

hkiang01:
For example, in a recent build, some libraries were updated to resolve 4 CVEs in the alpine image per `trivy image` compared to the latest available edge builds

### DIFF
--- a/docker/edge-alpine.Dockerfile
+++ b/docker/edge-alpine.Dockerfile
@@ -16,7 +16,7 @@ RUN curl -L -o /tmp/desktop-client.zip --header "Authorization: Bearer ${GITHUB_
 RUN unzip /tmp/desktop-client.zip -d /public
 
 FROM alpine:3.17 as prod
-RUN apk add --no-cache nodejs tini
+RUN apk update && apk upgrade && apk add --no-cache nodejs tini
 
 ARG USERNAME=actual
 ARG USER_UID=1001
--- a/docker/edge-ubuntu.Dockerfile
+++ b/docker/edge-ubuntu.Dockerfile
@@ -15,7 +15,7 @@ RUN curl -L -o /tmp/desktop-client.zip --header "Authorization: Bearer ${GITHUB_
 RUN unzip /tmp/desktop-client.zip -d /public
 
 FROM node:18-bullseye-slim as prod
-RUN apt-get update && apt-get install tini && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get upgrade && apt-get install tini && apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
 ARG USERNAME=actual
 ARG USER_UID=1001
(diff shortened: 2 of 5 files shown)


docker/edge-alpine.Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN curl -L -o /tmp/desktop-client.zip --header "Authorization: Bearer ${GITHUB_
 RUN unzip /tmp/desktop-client.zip -d /public
 
 FROM alpine:3.17 as prod
-RUN apk add --no-cache nodejs tini
+RUN apk update && apk upgrade && apk add --no-cache nodejs tini
```
MatissJanis, line 19: **💭 thought:** ‏(I'm not a docker expert so take this with a grain of salt)

Wouldn't this actually make the docker images _unstable_?

Before: every deployment of the docker app was predictable and equal. If it works on "my machine" it will work on yours.

After: every deployment is unique. On my machine I'll get `someDependency v1.2.3`, but on yours you might get `someDependency v1.2.9` a few hours later.

Which means.. your app might blow up unpredictably if an external dependency introduces a breaking change.

And believe me: these things do happen.. and more often than you'd expect.

--

What I would actually like to see is specific versions _hardcoded_ in the dockerfiles for dependencies. And then we slowly upgrade them across the board - **manually**.

hkiang01: Pinning versions then requires us to keep those pins updated. That's a pain.

Ideally stability would be ensured by comprehensive testing. This would give you the freedom to update subsequent layers with confidence. Does this project have comprehensive testing?

Absent that you could offload the responsibility of keeping OS packages up-to-date upstream so that they can balance security and stability of Node at large, and presumably the Node packages used by this project. While I expect it's rare for an updated version of an OS package to cause instability in this context, there is still a possibility it can happen, especially if some dependencies have a relatively complex build on install. If we're not comfortable with that, we should simply close (and not merge).

twk3: > Wouldn't this actually make the docker images unstable?

> Before: every deployment of the docker app was predictable and equal. If it works on "my machine" it will work on yours.

@MatissJanis this is at build time. Not at deploy time, so everyone using the same sha256 of the image will have the same dependencies. 

In addition, in terms of compatibility, it's not really different than the alpine3.17 image or node:18 image itself being updated upstream on us.

The downside is that any updates done in this way will add size to the image as a new layer. Rather than replace their existing size. 

@hkiang01 I don't think we get a lot of benefit from the update currently though. Our dockefiles are simple enough, that the node image is already getting these update on a monthly cadence. And the alpine one is tiny enough that  we would really only be getting openssl updates here, but a security release of openssl already typically results in a new alpine base image.

I guess what I am saying is that I don't think we will move to release faster than our base images will anyways, and because we aren't introducing our own extra base images, the important thing for us is to always be using base images that are still getting frequent security updates and support. 

TLDR: I think the change is fine, but unnecessary with our current base images. For example, at the time of writing this results in 0 updated packages. This is because both our base images have been updated within the last 2 weeks.

hkiang01: > I don't think we will move to release faster than our base images will anyways

SGTM, can close

PR #311, opened 2024-02-13
Fix the link page for gocardless

twk3:
Ensures the redirect page is auto-closed after linking

After we dropped the nordigen endpoint in https://github.com/actualbudget/actual-server/pull/293 this was no longer working.

(Closing the redirect page automatically will typically reveal the previous tab which is the waiting actual page, where you can then setup next steps)


MatissJanis (approved):
Thanks! I somehow missed this.

PR #315, opened 2024-02-21
Pending Transaction Support for SimpleFIN Integration

duplaja:
This pull request adds pending transaction support for SimpleFIN (for the accounts that support pending transactions).

It follows the lead of the GoCardless integration, by importing pending transactions initially as uncleared, and then clearing when the transactions post.

### DIFF
--- a/src/app-simplefin/app-simplefin.js
+++ b/src/app-simplefin/app-simplefin.js
@@ -96,11 +96,11 @@ app.post('/transactions', async (req, res) => {
 
       let dateToUse = 0;
 
-      if(trans.posted == 0) {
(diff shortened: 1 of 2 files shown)


youngcw:
Am I understanding this right? You are setting the booked field to each transaction to then leverage the existing logic on the client side to handle the pending?

duplaja:
That is correct. I didn't think I should re-invent the wheel, since we had that logic there.

I've done a full round of tests, to make sure it handles as expected:

Imports SimpleFIN transaction as pending (uncleared), then later once it is no longer pending, it clears in Actual.

youngcw (approved):

PR #316, opened 2024-02-28
feat: add SEB Privat bank

myhrmans:
Add SEB Private Bank integration to gocardless. Handle that SEB is sending the creditor name in additionalInfo.


MatissJanis (approved):
Thanks!

Issue #261, opened 2023-09-04
Flip amount option with GoCardless


parasdaryanani:
HSBC UK credit card transactions synced with GoCardless show a positive balance, which is actually the remaining credit limit. Other credit cards like Amex and Barclaycard correctly show a negative balance. Is there a way to use the "flip amount" feature with GoCardless account sync?

MatissJanis:
I would recommend first reaching out to GoCardless and letting them know of this inconsistency. They are pretty quick to respond and resolve issues.

Let me know what they say!

parasdaryanani:
@MatissJanis here is what they came back with:

`HSBC_HBUKGB4B` exposes 2 items under balances:

```json

{
    "balances": [
        {
            "balanceAmount": {
                "amount": "4183.61",
                "currency": "GBP"
            },
            "balanceType": "interimAvailable",
            "referenceDate": "2023-09-05"
        },
        {
            "balanceAmount": {
                "amount": "-316.39",
                "currency": "GBP"
            },
            "balanceType": "interimBooked",
            "referenceDate": "2023-09-05"
        }
    ]
}
```

It looks like Actual is using `interimAvailable` for current balance instead of `interimBooked` for cards.

MatissJanis:
Interesting. Yea, we might not be properly handling credit cards. Or perhaps just your banks credit cards.. I don't have a CC, so can't check what I would get on my end sadly.

The bank adapters can be found here: https://github.com/actualbudget/actual-server/tree/master/src/app-gocardless/banks

Worst case: you can add a custom adapter specifically for your bank.

dylmye:
I can confirm this is the same with Santander UK credit cards.

parasdaryanani:
I'll make some time to play around with this and see if I can implement a generic solution that will check for interimBooked, specifically for credit cards. Otherwise, custom bank adapter it is.

jakoblover:
I am having the same issue with `SEB_KORT_AB_NO_SKHSFI21`, where purchases done with the card are positive and payments to the card are negative. I contacted GoCardless and this is the response I got:

> Thanks for reaching out! 
According to the bank - this is the expected behaviour for these transactions as they have been made from a credit card account and the amounts are referencing your credit limit balance. 
If a debit card account would be connected the transactions would show up in the way you are expecting as then it would be referencing the debit balance.

jzahedieh:
[...]

PR #325, opened 2024-03-20
feat: add 'SEB Kort Bank AB' credit card support

walleb:
Add custom bank adapter for 'SEB Kort Bank AB' to properly sync credit card transactions.

* Flip transaction amount so purchases are listed as payments (related to #261 )
* Handle that SEB is sending the creditor name in additionalInfo (copied from #316)
* Balance calculation based on 'expected' balance type


MatissJanis:
👋 Just checking if you're still working here or if this is ready for review. The PR is in "draft".

walleb:
Yes, still working on it. I put it into "draft" as a realized that I need to verify the balance calculation. I suspect that it now shows a positive amount when it should be negative. Tricky to get it right for credit card accounts...

walleb:
Yes, still working on it. I put it into "draft" as a realized that I need to verify the balance calculation. Balance calculation should be OK now for total, cleared and uncleared balances. It works properly when booking a payment of credit card bill with payee as one of the other accounts.

MatissJanis (approved):
Thanks

PR #328, opened 2024-03-26
:bug: (asn-bank) add special ASN Bank integration to patch whitespace issue

MatissJanis:
Patch for: https://github.com/actualbudget/actual/issues/2495

### DIFF
--- a/src/app-gocardless/bank-factory.js
+++ b/src/app-gocardless/bank-factory.js
@@ -1,4 +1,5 @@
 import AmericanExpressAesudef1 from './banks/american-express-aesudef1.js';
+import AsnBank from './banks/asn-bank.js';
 import IngIngddeff from './banks/ing-ingddeff.js';
 import IngPlIngbplpw from './banks/ing-pl-ingbplpw.js';
 import IntegrationBank from './banks/integration-bank.js';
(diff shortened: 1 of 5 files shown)


kyrias:
I wonder if we should really perform this unconditionally though, since I don't think there's any valid case in which we'd want to accept such a transaction ID.

MatissJanis:
> I wonder if we should really perform this unconditionally though, since I don't think there's any valid case in which we'd want to accept such a transaction ID.

Generally I agree with you. But see the conversation here: https://github.com/actualbudget/actual/issues/2495

The author is arguing that empty transaction IDs are expected. And since I don't have access to this bank to actually test things - I'll take his word for this.

We can always change this in the future

kyrias:
Well that's quoting yourself and not the author.  :p

But well, fair enough.  Though I'd argue that if we're worried about this being a breaking change at all then it's still possible for this to be a breaking change for other kinds of transactions for the same bank.  Arguably this should instead be reported to GoCardless to be fixed up by them removing the transaction ID if this case happens.

MatissJanis:
> [..] as an optional but _unique_ identifier [..]

Would you mind linking to the docs where they state this? If this is the case, then it does make much more sense to report it to GoCardless and get it patched on their end.

kyrias:
> | transactionId | Max256Text | Optional | Unique transaction identifier given by financial institution |
— https://developer.gocardless.com/bank-account-data/transactions

MatissJanis:
Thanks! I'll revert this back to OP in the github issue and close this PR for now. Lets see if we can get GoCardless to patch this on their end.

PR #333, opened 2024-04-07
Add ACTUAL_PORT environment variable to docker-compose.yml

psybers:
It can be difficult to see how to bind Actual server to a different port when using `docker compose`.  This adds a (commented out) environment variable for `ACTUAL_PORT` to make that more obvious to users.


MatissJanis (approved):
Makes sense. Thanks!

Issue #321, opened 2024-03-08
[Bug]: Link Your Bank (Go Cardless > Ireland > Permanent TSB) results in 500 Server Error - "An error occurred while linking your account, sorry!"


getoffmyplane:
### Verified issue does not already exist?

- [X] I have searched and found no existing issue

### What happened?

The latest version of Actual Budget (which updated the GoCardless API endpoints to the new ones) require my accounts to be relinked via GoCardless.

I use two banks in Ireland - Allied Irish Banks and Permanent TSB.

Allied Irish Banks works fine.

Trying to link my bank account (Country Ireland, Bank Permanent TSB) via GoCardless (on the "Link Your Bank" popup) causes the following error:
500 Server Error - "An error occurred while linking your account, sorry!"

To note:

This error was there before the API endpoints were changed in the latest version of Actual Budget but I was able to work around it by selecting a different bank from the "Link Your Bank" popup, and then updating the institution id in the url on the subsequent browser screen (the Gocardless auth screen).

I got the institution ID from https://docs.google.com/spreadsheets/d/1EZ5n7QDGaRIot5M86dwqd5UFSGEDTeTRzEq3D9uEDkM (which is linked to from the gocardless support section).

Unfortunately, that workaround no longer works (the tab just closes when you press the "agree and accept" button after you change the institution id part of the url).

I checked the response from get-banks (fired by the "Link Your Bank" popup) with Firefox inspect and the response seems to match what is in the spreadsheet

Actual Budget is running in Docker on Ubuntu Server.
I've tried Firefox and Chrome on Windows 11, Fedora 39 & Debian 12 - all return same error.

If I can provide any more info, please let me know.

Thanks!

### What error did you receive?

500 Server Error - "An error occurred while linking your account, sorry!"

### Where are you hosting Actual?

Docker

### What browsers are you seeing the problem on?

Firefox, Chrome

### Operating System

Windows 11

getoffmyplane:
Hi again,

I think this may just be that max_historical_days is set to 90 when the max the bank supports, according to GoCardless' spreadsheet, is 89 (go to https://docs.google.com/spreadsheets/d/1EZ5n7QDGaRIot5M86dwqd5UFSGEDTeTRzEq3D9uEDkM, then search "Permanent TSB", then see column D).

Would really appreciate a steer (a bit out of my depth) - if there's somewhere i can update this myself, more than happy to do it.

Thanks!

mickotoole02:
[...]

PR #334, opened 2024-04-08
Allow setting accessValidForDays and maxHistoricalDays per bank

kyrias:
Right now this shouldn't actually change anything because the Actual will only try to fetch at most 90 days worth of transactions anyway.  I'll file a PR for those changes if this is approved.

I set the values for the existing integrations based on [this spreadsheet](https://docs.google.com/spreadsheets/d/1EZ5n7QDGaRIot5M86dwqd5UFSGEDTeTRzEq3D9uEDkM/edit#gid=976380583) and [these](https://nordigen.zendesk.com/hc/en-gb/articles/13239212055581-EEA-180-day-access) [two](https://nordigen.zendesk.com/hc/en-gb/articles/6760902653085-Extended-history-and-continuous-access-edge-cases) documentation pages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "express-response-size": "^0.0.3",
     "jws": "^4.0.0",
     "migrate": "^2.0.1",
-    "nordigen-node": "^1.3.0",
+    "nordigen-node": "^1.4.0",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
--- a/src/app-gocardless/app-gocardless.js
+++ b/src/app-gocardless/app-gocardless.js
@@ -40,11 +40,10 @@ app.post('/status', async (req, res) => {
 app.post(
   '/create-web-token',
   handleError(async (req, res) => {
-    const { accessValidForDays, institutionId } = req.body;
+    const { institutionId } = req.body;
     const { origin } = req.headers;
 
     const { link, requisitionId } = await goCardlessService.createRequisition({
-      accessValidForDays,
       institutionId,
       host: origin,
     });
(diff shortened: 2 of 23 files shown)


kyrias:
Actually, there's something strange going on here.  I just looked at the requisitions that Actual had previously created and the one from December for one of my banks already had an `accessValidForDays` of 90, so I think that we're using the API wrong and it's not getting set correctly...

kyrias:
Ah, seems it was a bug in the nordigen module: https://github.com/nordigen/nordigen-node/commit/5beefda1680ddaf4547c8c3e6792f461fd13a7bd

But uh.  We're supposedly using 1.3.0 but the commit that updated the version to 1.3.0 (https://github.com/nordigen/nordigen-node/commit/05d261c95a28c74b7ac1a71e3f4fd8b4457a9f4f) happened *after* that fix, but the fix doesn't exist in the npm 1.3.0 package?  :/

kyrias:
Yeah, upgrading to 1.4.0 has it, so it seems like 1.3.0 on NPM doesn't at all correspond to their commits that bump the version to 1.3.0, meaning that there is no (public at least) commit of the project which corresponds to the 1.3.0 that's published on NPM.  :upside_down_face: 

I guess I'll update the dependency as well then.

MatissJanis:
Has anyone actually reported this issue to GoCardless? I'd be interested to hear back from them (since this causes 500s on their end) before jumping to solutions on our end.

kyrias:
@MatissJanis What issue?

MatissJanis:
https://github.com/actualbudget/actual-server/issues/321

kyrias:
That issue isn't why I did this, but the issue is also misleading because the GoCardless does not return a 500 error, it's the Actual server that does.  GoCardless' API returns a correct error saying exactly what's wrong with the request.

kyrias:
Though I'd also missed that they actually serve the maximum allowed `max_historical_days` value in their API, so this should be changed to get the correct value from there instead.  Annoyingly they do not serve the maximum allowed `access_valid_for_days` though.

MatissJanis (approved):
Thanks!

PR #342, opened 2024-04-17
Make /admin/login return a descriptive error when no password is provided

### LINKED ISSUE #308: [Bug]: Trying to login without a body causes internal server error

### What happened?

Internal server error on endpoint `/account/login` if no payload is provided. Can be easily reproduced by spinning a local version of Actual via docker and doing a request with curl:

```
~ ➤ curl -X POST http://localhost:5006/account/login
{"status":"error","reason":"internal-error"}%
~ ➤ curl -X POST -I http://localhost:5006/account/login
HTTP/2 500
access-control-allow-origin: *
content-type: application/json; charset=utf-8
date: Sun, 04 Feb 2024 14:07:28 GMT
etag: W/"2c-eqop7Vf+ETHUweBQ0x2nOktTJvQ"
ratelimit-limit: 500
ratelimit-remaining: 499
ratelimit-reset: 56
x-powered-by: Express
content-length: 44
```

The response came as `<500: Internal server error>`. I would either expect the default "ok" with a valid (or null) token, or a `<400 Bad Request>`. Here is the exact error log on the server side:

```
ERROR Error: data and hash arguments required
    at Module.compareSync (/app/node_modules/bcrypt/bcrypt.js:166:15)
    at login (file:///app/src/account-db.js:54:33)
    at file:///app/src/app-account.js:41:15
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at next (/app/node_modules/express/lib/router/route.js:144:13)
    at Route.dispatch (/app/node_modules/express/lib/router/route.js:114:3)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at /app/node_modules/express/lib/router/index.js:284:15
    at Function.process_params (/app/node_modules/express/lib/router/index.js:346:12)
    at next (/app/node_modules/express/lib/router/index.js:280:10)
```

Docker, version of actual is 24.2.0

_No response_

### Operating System

Windows 10

matt-fidd:
Fixes #308 

This change makes `/account/login` consistent with other non-authenticated endpoints and adds a more descriptive error message when no password is provided in the request body.

Before
```sh
% curl -X POST http://localhost:5006/account/login                                                                                                                                                                                                                   
{"status":"error","reason":"internal-error"}
```

After
```sh
% curl -X POST http://localhost:5006/account/login                                                                                                                                                                                                                   
{"status":"error","reason":"invalid-password"}
```

### DIFF
--- a/src/account-db.js
+++ b/src/account-db.js
@@ -49,7 +49,7 @@ export function bootstrap(password) {
 }
 
 export function login(password) {
-  if (password == null || password === '') {
+  if (password === undefined || password === '') {
     return { error: 'invalid-password' };
   }
 
--- a/src/account-db.js
+++ b/src/account-db.js
@@ -49,8 +49,13 @@ export function bootstrap(password) {
 }
 
 export function login(password) {
+  if (password == null || password === '') {
+    return { error: 'invalid-password' };
+  }
+
   let accountDb = getAccountDb();
   let row = accountDb.first('SELECT * FROM auth');
+
   let confirmed = row && bcrypt.compareSync(password, row.password);
 
   if (confirmed) {
@@ -59,7 +64,7 @@ export function login(password) {
     // "session" that times out after a long time or something, and
     // maybe each device has a different token
     let row = accountDb.first('SELECT * FROM sessions');
-    return row.token;
+    return { token: row.token };
   } else {
     return null;
   }
--- a/src/app-account.js
+++ b/src/app-account.js
@@ -38,8 +38,14 @@ app.post('/bootstrap', (req, res) => {
 });
 
 app.post('/login', (req, res) => {
-  let token = login(req.body.password);
-  res.send({ status: 'ok', data: { token } });
+  let { error, token } = login(req.body.password);
+
+  if (error) {
+    res.status(400).send({ status: 'error', reason: error });
+    return;
+  } else {
+    res.send({ status: 'ok', data: { token } });
+  }
 });
 
 app.post('/change-password', (req, res) => {
--- a/src/app-account.js
+++ b/src/app-account.js
@@ -43,9 +43,9 @@ app.post('/login', (req, res) => {
   if (error) {
     res.status(400).send({ status: 'error', reason: error });
     return;
-  } else {
-    res.send({ status: 'ok', data: { token } });
   }
+
+  res.send({ status: 'ok', data: { token } });
 });
 
 app.post('/change-password', (req, res) => {
--- a/upcoming-release-notes/342.md
+++ b/upcoming-release-notes/342.md
@@ -1,0 +1,6 @@
+---
+category: Bugfix
+authors: [matt-fidd]
+---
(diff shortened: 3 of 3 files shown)


src/account-db.js
```diff
@@ -49,8 +49,13 @@ export function bootstrap(password) {
 }
 
 export function login(password) {
+  if (password == null || password === '') {
```
MatissJanis, line 52: **💬 suggestion:** ‏would you mind following strict JS here?
```suggestion
  if (password === null || password === '') {
```

src/app-account.js
```diff
@@ -38,8 +38,14 @@ app.post('/bootstrap', (req, res) => {
...
-  let token = login(req.body.password);
-  res.send({ status: 'ok', data: { token } });
+  let { error, token } = login(req.body.password);
+
+  if (error) {
+    res.status(400).send({ status: 'error', reason: error });
+    return;
+  } else {
+    res.send({ status: 'ok', data: { token } });
+  }
```
MatissJanis, line 48: **💬 suggestion:** ‏unnecessary `else` statement

```suggestion

  res.send({ status: 'ok', data: { token } });
```

MatissJanis (changes requested):
Thanks for working on this! Just two small change requests, but other than that this looks great!

src/account-db.js
matt-fidd: I think that it actually comes through as `undefined` so the weak comparison to `null` was required here to pick out null-ish. I can change it to be `password === undefined` if you'd prefer but I was trying to keep it consistent with similar checks in the same file.

MatissJanis: We could do both `password === undefined || password === null`. Eventually we will refactor the entire repo to be strict.. but for now we can focus only on the new code :)

src/app-account.js
matt-fidd: Agreed, but this was kept consistent with the current approach in this file. Happy to amend this one if you'd like, and I can fix other instances in this file while I'm there too.

```js
app.post('/bootstrap', (req, res) => {                                          
  let { error, token } = bootstrap(req.body.password);                          
                                                                                
  if (error) {                                                                  
    res.status(400).send({ status: 'error', reason: error });                   
    return;                                                                     
  } else {                                                                      
    res.send({ status: 'ok', data: { token } });                                
  }                                                                             
});
```

MatissJanis: Yea this repo isn't the cleanest.. :) 

I primarily focus on the new additions and try to get those as clean as possible. So wouldn't want to block you on that. But if you have time to also amend the existing code - that would be amazing!

MatissJanis (approved):

src/app-account.js
matt-fidd: Of course! I'd be happy to run through and do those others if you give me a couple of minutes :)

PR #343, opened 2024-04-19
Refactor to add strict type comparisons and remove unnecessary else blocks

matt-fidd:


src/account-db.js
```diff
@@ -58,31 +59,33 @@ export function login(password) {
...
-    // Right now, tokens are permanent and there's just one in the
-    // system. In the future this should probably evolve to be a
-    // "session" that times out after a long time or something, and
-    // maybe each device has a different token
-    let row = accountDb.first('SELECT * FROM sessions');
-    return { token: row.token };
-  } else {
-    return null;
+  if (!confirmed) {
+    return { error: 'invalid-password' };
```
matt-fidd, line 63: This is the only functional change in this PR, if the password is incorrect the server will return `invalid-password`, not `null`.

MatissJanis (approved):
Thanks!

PR #345, opened 2024-04-22
Add backup date field for GoCardless transactions with bank BNP_BE_GEBABEBB

### LINKED ISSUE #344: [Bug]: GoCardLess Sync: Error: 'date' is required when adding a transaction

### What happened?

When syncing an account via GoCardLess, it failed with an error message suggesting to contact support. The expectation is that the sync completes.

This issue happens for bank BNP_BE_GEBABEBB.

This issue has been discussed on Discord. 

In the console, the following transactions are visible, one of which is unbooked and lacks a date: 
image

A screenshot from a few days later (when sync completes) shows that bookingDate and valueDate are close but not identical: 
image

Docker

Safari

### Operating System

Mac OSX

matt-fidd:
Fixes #344 

As discussed on Discord, pending transactions do not come through from this bank with a valueDate set, only bookingDate.

However bookingDate is close enough to valueDate that the transaction will deduplicate when booked so we will use bookingDate until valueDate is available.

### DIFF
--- a/src/app-gocardless/banks/bnp-be-gebabebb.js
+++ b/src/app-gocardless/banks/bnp-be-gebabebb.js
@@ -73,7 +73,7 @@ export default {
 
     return {
       ...transaction,
-      date: transaction.valueDate,
+      date: transaction.valueDate || transaction.bookingDate,
     };
   },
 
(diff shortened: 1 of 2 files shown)


MatissJanis (approved):
Great!

PR #346, opened 2024-04-29
Add gocardless support for Sparkasse Karlsruhe (Germany)

### LINKED ISSUE #331: [Bug]: "ultimateCreditor" information from GoCardless is not used

### What happened?

When importing transactions from GoCardless, the information in the Payee/Notes field is often lacking. With my bank, Sparkasse Karlsruhe, "Notes" is never populated at all. The "Payee" is often very unhelpful, since many businesses use a payment provider for their card payments. Then, only the provider or a generic description ends up in "Payee". This seems to be the `creditorName` field. Often, a `ultimateCreditor` field is available in GoCardless data. If this field is available and populated, it should be used instead of the `creditorName`. This way, more sensible information can reach the user.

Example of a `creditorName` i often see is "Lastschriftzahlung" (=debit charge). Not helpful - the `ultimateCreditor` would then include "Some Nice Cafe where i bought a cake".

_No response_

Docker

_No response_

### Operating System

None

Nebukadneza:
Adds a gocardless formatter, to properly format transaction information. Previously, additional remittance information was not supplied at all, and some creditor names were not set. This made recognizing transactions hard for users.

Now, we pick out the best remittanceInformation that we find, and use a helpful order of possible data from the transaction as the creditorName.

Fixes #331

### DIFF
--- a/src/app-gocardless/bank-factory.js
+++ b/src/app-gocardless/bank-factory.js
@@ -12,6 +12,7 @@ import SEBPrivat from './banks/seb-privat.js';
 import SandboxfinanceSfin0000 from './banks/sandboxfinance-sfin0000.js';
 import SparNordSpNoDK22 from './banks/sparnord-spnodk22.js';
 import SpkMarburgBiedenkopfHeladef1mar from './banks/spk-marburg-biedenkopf-heladef1mar.js';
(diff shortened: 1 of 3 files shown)


MatissJanis (approved):
Thanks!

PR #349, opened 2024-05-01
fill in creditor name for BNP bank transactions

vojeroen:
Ensure payee names don't contain transactional information when pulling in transactions from BNP bank with GoCardless.


vojeroen:
Also, it looks like this change poses no problems when re-processing transactions that have already been synced before, but now with a different creditorName. I suppose the GoCardless sync code is already intelligent enough to filter that out.

MatissJanis:
@CharlieMK @feliciaan 
If either of you are around - would you be open to check this PR? Both of you worked on this bank adapter and I wouldn't want things to break for you.

MatissJanis (approved):

PR #350, opened 2024-05-02
Extend bank adapter for SEB Kort

jakoblover:
`SEB_KORT_AB_NO_SKHSFI21` behaves like `SEB_KORT_AB_SE_SKHSFI21`. Added `SEB_KORT_AB_NO_SKHSFI21` to the list of institutions in the bank adapter for SEB Kort. 

~~I'm not able to get the correct starting balance. It seems like the transactions array does not include the non-invoiced transactions, only the booked transactions. In the overview of transactions, however, the non-invoiced transactions are listed. The starting balance is missing the amount equivalent of the non-invoiced transactions. Could use some help here.~~

Seems like I am able to get the correct starting balance after all. I had to connect the accont to gocardless again for it to recalculate the starting balance.

### DIFF
--- a/src/app-gocardless/banks/seb-kort-bank-ab.js
+++ b/src/app-gocardless/banks/seb-kort-bank-ab.js
@@ -6,7 +6,7 @@ import {
 
 /** @type {import('./bank.interface.js').IBank} */
 export default {
-  institutionIds: ['SEB_KORT_AB_SE_SKHSFI21'],
+  institutionIds: ['SEB_KORT_AB_NO_SKHSFI21','SEB_KORT_AB_SE_SKHSFI21'],
 
   accessValidForDays: 180,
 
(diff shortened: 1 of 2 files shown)


MatissJanis:
@walleb if you are still around - would you mind taking a peak at this PR? It _slightly_ changes the existing SEB bank mapper. I wouldn't want things to break for you.

On my end this looks good to go 👍

walleb:
> @walleb if you are still around - would you mind taking a peak at this PR? It _slightly_ changes the existing SEB bank mapper. I wouldn't want things to break for you.

Looks good to me. Not sure how I missed the "nonInvoiced" balance type when I tested my PR. Maybe I didn't had any such transactions at that time. 👍

MatissJanis (approved):
Thanks!

PR #353, opened 2024-05-06
Move bank sync payee name normalisation from `actual` to `actual-server`

### LINKED ISSUE #348: [Bug]: GoCardless inconsistent use of creditorName and debtorName fields

### What happened?

It appears that GoCardless are inconsistent across multiple banks with their use of creditor vs debtor name. Whilst I agree that it really is GoCardless' bug to fix, it seems that it effects multiple banks from my limited testing (and I've only got access to a few providers!).

This results in us falling back to unstructured information for Payee which normally includes dates and one-off transaction information.

I propose that we use the "correct" field first, and if it's not present fallback to the "incorrect".

The proposed change could either take place in `src/app-gocardless/banks/integration-bank.js` in this repository where the below could be added to the `normalizeTransaction` function. (I prefer this option)

```diff
+    if (transaction.creditorName && !transaction.debtorName) {
+      transaction.debtorName = transaction.creditorName;
+    } else if (transaction.debtorName && !transaction.creditorName) {
+      transaction.creditorName = transaction.debtorName;
+    }
```

Or could be applied in the `actual` repository in `packages/loot-core/src/server/accounts/sync.ts` in the `normalizeBankSyncTransactions` function.

matt-fidd:
This PR fixes #348 and takes the approach discussed to add a new `payeeName` field to the transaction returned to `actual`.

It is designed to fix any inconsistencies in data returned from GoCardless with regards to the variety of payee fields that are used, and should result in the "best chance" of a human readable and non-transaction specific payee if it is available.

I have tested these with the bank that I originally spotted the issue with (Santander) and payee names are now correctly created in `actual`. Banks that I can test with that already worked as expected have not regressed and payees are still reflected correctly.

I can not test the custom bank handlers but all I have done to them is make sure they now return a `payeeName` as this will be required by `actual` to normalise the transactions on the client side.

I've also had to copy the `title` util that `actual` used to format the payee names.

Complimentary PR in Actual is: actualbudget/actual#2721


psybers:
As mentioned in https://github.com/actualbudget/actual/pull/2721 maybe the field can be named `payee` instead of `payeeName`?  That way it should work with SimpleFIN too.

psybers:
And I do wonder, does this function need called on the SimpleFIN side too?

matt-fidd:
Now that I'm thinking about this again, I'm not sure this is the best approach for gocardless. Previously each custom handler had this payee standardisation done regardless, but this change puts the onus on the contributer to make sure the field is included in the handler.

Maybe I could move this into `gocardless-service` to retain the behaviour from before?

https://github.com/actualbudget/actual-server/blob/db6b4e4800448625fe0be969938b7673c3badbae/src/app-gocardless/services/gocardless-service.js#L448

matt-fidd:
> Now that I'm thinking about this again, I'm not sure this is the best approach for gocardless. Previously each custom handler had this payee standardisation done regardless, but this change puts the onus on the contributer to make sure the field is included in the handler.
> 
> Maybe I could move the formatPayeeName call into `gocardless-service` to retain the behaviour from before and run on all transactions by default?
> 
> https://github.com/actualbudget/actual-server/blob/db6b4e4800448625fe0be969938b7673c3badbae/src/app-gocardless/services/gocardless-service.js#L448
> 
> Any thoughts?

Actually I reckon the above is the better way to go, it retains the behaviour from before and simplifies bank handlers. Will refactor now.

src/app-gocardless/services/gocardless-service.js
```diff
@@ -211,6 +212,20 @@ export const goCardlessService = {
       goCardlessService.getBalances(accountId),
     ]);
 
+    transactions.transactions.booked = transactions.transactions?.booked?.map(
```
psybers, line 215: Won't this crash?

If `transactions.transactions` is undefined, the right hand side will eventually return an undefined value.  But you then try to set that value to `transactions.transactions.booked` which should give an undefined error when accessing `transactions.transactions`.

I think you need to guard this block with `if (transactions.transactions)`.

psybers (changes requested):

src/app-gocardless/services/gocardless-service.js
matt-fidd: You're right, I was copying the sort logic from below but looking into the `getTransactions` function, it doesn't look like this could ever be undefined.

https://github.com/actualbudget/actual-server/blob/db6b4e4800448625fe0be969938b7673c3badbae/src/app-gocardless/services/gocardless-service.js#L458-L465

Would you prefer i guard this or remove the optional chaining operator entirely?

psybers: I don't know that API but if it is never undefined, then dropping the `?`'s makes sense.

Does it always return an object with `booked` and `pending` set, even if there are no transactions of those kind?

matt-fidd: Just checked, booked and pending will always be an array and will be left empty if there are no transactions so the chaining here isn't required. Removed it now

src/util/payee-name.js
```diff
@@ -0,0 +1,42 @@
...
+
+  // use the correct fields for the amount if it was found
+  // if not, use whatever we can find
+  name =
+    name ||
+    trans.debtorName ||
+    trans.creditorName ||
+    trans.remittanceInformationUnstructured ||
+    (trans.remittanceInformationUnstructuredArray || []).join(', ') ||
+    trans.additionalInformation;
```
psybers, line 29: Since the if/else above sets `name` to something (possibly), you can avoid the assignment here and simplify it quite a bit in that case it found a name already.

```suggestion
  if (!name) {
    name =
      trans.remittanceInformationUnstructured ||
      (trans.remittanceInformationUnstructuredArray || []).join(', ') ||
      trans.additionalInformation;
  }
```

src/util/payee-name.js
```diff
@@ -0,0 +1,42 @@
...
+  let account;
+  if (trans.amount > 0 || Object.is(Number(trans.amount), 0)) {
+    name = trans.debtorName;
+    account = trans.debtorAccount;
+  } else {
+    name = trans.creditorName;
+    account = trans.creditorAccount;
+  }
+
+  // use the correct fields for the amount if it was found
```
psybers, line 21: The word `amount` in this comment seems to be wrong.  Maybe it should be `name`?

psybers (changes requested):

src/util/payee-name.js
matt-fidd: This changes the logic, the key functional part of this change is to allow creditorName to be used even if GoCardless /should/ be using debtorName instead and vice versa.

Happy to put the whole assignment into an if and drop the first `name || ` line if that aligns more closely with the style in this project?

psybers: That is curious logic!  I guess if that is what needs to happen for GoCardless, the existing code is probably fine.  I am not sure if there is a style preference between using an `if` to avoid `name = name` or leave it as is.

src/app-gocardless/services/gocardless-service.js
```diff
@@ -211,6 +212,20 @@ export const goCardlessService = {
...
+        payeeName: formatPayeeName(t),
+      }),
+    );
+
+    transactions.transactions.pending = transactions.transactions.pending.map(
+      (t) => ({
+        ...t,
+        payeeName: formatPayeeName(t),
+      }),
+    );
```
MatissJanis, line 227: **💬 suggestion:** ‏instead of doing this uniformly for ALL banks the same way - I think we should allow the bank-providers to do their own payee normalisation (if they wish to).

So my suggestion here is to move this payee name normalisation inside [here](https://github.com/actualbudget/actual-server/blob/1da9c821ee858a38f9bbc11d4949a398a44c989a/src/app-gocardless/banks/integration-bank.js#L48).

Other than that - LGTM!

matt-fidd: That was my initial approach but I turned away from it because this way replicates the behaviour prior to this change
If needed creditorName/debtorName can be altered in the bank handler to change the payee name, but I think that the default behaviour to fall back on other fields and append the IBAN when present is best retained.

Let me know what you reckon, happy to change if you disagree.

MatissJanis: We can still keep the same behavior as before. But just have the normalization logic in a different location (the generic bank handler + the overrides). 

IMO that would be more future proof solution since it would allow to change this logic on a per-bank basis.

src/app-gocardless/banks/nationwide-naiagb21.js
```diff
@@ -23,8 +23,6 @@ export default {
       transaction.bookingDate = useDate.toISOString().slice(0, 10);
     }
 
-    console.log(transaction);
-
```
MatissJanis, line 27: **👏 praise:** ‏thanks!

MatissJanis (approved):

MatissJanis:
Before merging this in: @psybers do you have any other feedback here? Or good to go on your end too?

psybers:
@MatissJanis I think for now this is fine. I am a bit worried about having `payeeName: formatPayee(transaction)` all over the place, and specifically that it might get missed on new bank overrides.  But we can probably find a cleaner way to handle that and refactor it down the road.

PR #358, opened 2024-05-16
Add gocardless support for Berliner Sparkasse (Germany)

Peccadilloz:
Similar to PR #346 this adds a gocardless formatter to properly format transaction information. The additional remittance information was not picked up by actual as this bank transmits it as remittanceInformationStructured instead of remittanceInformationUnstructured. 

I have tested this successfully with my bank. The "Notes" section is now being filled with the additional remittance information.

This fixes my original issue in the actual repo  [#2759](https://github.com/actualbudget/actual/issues/2759)


src/app-gocardless/banks/berliner_sparkasse_beladebexxx.js
```diff
@@ -0,0 +1,98 @@
...
+    return {
+      ...transaction,
+      creditorName: usefulCreditorName,
+      remittanceInformationUnstructured: remittanceInformationUnstructured,
+      date: transaction.bookingDate || transaction.valueDate,
+    };
+  },
+
+  sortTransactions(transactions = []) {
+    return sortByBookingDateOrValueDate(transactions);
```
psybers, line 78: This seems like the default case.  Could you import and use the fallback version instead?

See this file for an example: https://github.com/actualbudget/actual-server/blob/master/src/app-gocardless/banks/belfius_gkccbebb.js

psybers (changes requested):

psybers (changes requested):
Can you add a test case?  See https://github.com/actualbudget/actual-server/blob/master/src/app-gocardless/banks/tests/belfius_gkccbebb.spec.js for an example.

psybers:

psybers:

Peccadilloz:
Sorry for being inactive this last month. I'll be back home this weekend and will pick this up.

Peccadilloz:
> Thanks for working on this! Would you mind updating the PR to use prop spreading for `Fallback`? Here's an example bank provider:
> 
> https://github.com/actualbudget/actual-server/blob/aa22e6951dd90fc03e086593f001d0a002b6e324/src/app-gocardless/banks/american-express-aesudef1.js#L7
> 
> We recently merged some changes to simplify them.
> 
> Happy to merge this PR once that's done + the conflict is solved!

I have updated this PR to include the requested changes and resolved the mentioned conflict. Any chance this can be included in the upcoming update still? 
Best regards! 
Peccadilloz

MatissJanis (approved):
Thanks!

You'll be able to use this change via the next `edge` tag (tomorrow). Sorry for not getting it in the current release. It barely missed the cut.

Peccadilloz:
> Thanks!
> 
> You'll be able to use this change via the next `edge` tag (tomorrow). Sorry for not getting it in the current release. It barely missed the cut.

No worries, I knew I was late for the release. 
Thanks for the quick merge!

tennox:
@Peccadilloz do you think I can just add the IDfor Mittelbrandenburgische Sparkasse and it'll work? 🤔
I'm also missing the remittance information as a note

DirgoSalga:
I am also having the same trouble with my bank. It is the Rheinhessen Sparkasse (that is the official name), the name under GoCardless was: Sparkasse Worms-Alzey-Ried (BIC: MALADE51WOR).

PR #359, opened 2024-05-18
Grab payee name for Abanca gocardless integration

daca11:
Current GoCardless integration with Abanca is not able to get Payee name because `creditorName` and `debtorName` are empty.
That information is actually in the `remittanceInformationStructured`. This fix adds that information to `creditorName` and `debtorName` so the payee name is correctly imported into the transaction correctly.


psybers (changes requested):
Can you add a test case? See https://github.com/actualbudget/actual-server/blob/master/src/app-gocardless/banks/tests/belfius_gkccbebb.spec.js for an example.

psybers:

psybers:

MatissJanis:
Thanks for working on this! Would you mind updating the PR to use prop spreading for `Fallback`? Here's an example bank provider: https://github.com/actualbudget/actual-server/blob/aa22e6951dd90fc03e086593f001d0a002b6e324/src/app-gocardless/banks/american-express-aesudef1.js#L7

We recently merged some changes to simplify them.

Happy to merge this PR once that's done + the conflict is solved!

daca11:
Added a test and the ...Fallback line.
Sorry for the trial-and-error commits, as a DevOps I never worked with ts. I hope the test is correct, I just based my work on the existing files 😄

src/app-gocardless/banks/abanca-caglesmm.js
```diff
@@ -0,0 +1,32 @@
...
+    };
+  },
+
+  sortTransactions(transactions = []) {
+    return Fallback.sortTransactions(transactions);
+  },
+
+  calculateStartingBalance(sortedTransactions = [], balances = []) {
+    return Fallback.calculateStartingBalance(sortedTransactions, balances);
+  },
```
MatissJanis, line 31: Can remove these. The prop spreading on line 5 takes care of adding them.

```suggestion
```

src/app-gocardless/banks/abanca-caglesmm.js
```diff
@@ -0,0 +1,32 @@
...
+  ...Fallback,
+
+  institutionIds: ['ABANCA_CAGLESMM'],
+
+  accessValidForDays: 180,
+
+  normalizeAccount(account) {
+    return Fallback.normalizeAccount(account);
+  },
+
```
MatissJanis, line 14:
```suggestion
```

MatissJanis (changes requested):

MatissJanis (approved):
Thanks!

PR #360, opened 2024-05-22
Add bank handler for `VIRGIN_NRNBGB22` (Virgin Money)

matt-fidd:
This change adds handling in for Virgin Money (VIRGIN_NRNBGB22). They don't return `creditorName` or `debtorName` through GoCardless so all transactions use the unprocessed `remittanceInformationUnstructured` field which contains extra details that make creating rules harder.

### DIFF
--- a/src/app-gocardless/banks/virgin_nrnbgb22.js
+++ b/src/app-gocardless/banks/virgin_nrnbgb22.js
@@ -12,7 +12,7 @@ export default {
 
   normalizeTransaction(transaction, booked) {
     const transferPrefixes = ['MOB', 'FPS'];
-    const methodRegex = /(Card|WLT)\s\d*/;
+    const methodRegex = /^(Card|WLT)\s\d+/;
 
     const parts = transaction.remittanceInformationUnstructured.split(', ');
 
--- a/src/app-gocardless/banks/virgin_nrnbgb22.js
+++ b/src/app-gocardless/banks/virgin_nrnbgb22.js
@@ -1,0 +1,44 @@
+import Fallback from './integration-bank.js';
+
+/** @type {import('./bank.interface.js').IBank} */
+export default {
+  institutionIds: ['VIRGIN_NRNBGB22'],
+
+  accessValidForDays: 90,
+
+  normalizeAccount(account) {
+    return Fallback.normalizeAccount(account);
+  },
+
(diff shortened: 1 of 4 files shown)


psybers (changes requested):
Can you add a test case? See https://github.com/actualbudget/actual-server/blob/master/src/app-gocardless/banks/tests/belfius_gkccbebb.spec.js for an example.

src/app-gocardless/banks/virgin_nrnbgb22.js
```diff
@@ -0,0 +1,49 @@
...
+
+  accessValidForDays: 90,
+
+  normalizeAccount(account) {
+    return Fallback.normalizeAccount(account);
+  },
+
+  normalizeTransaction(transaction, booked) {
+    const transferPrefixes = ['MOB', 'FPS'];
+    const methodRegex = /(Card|WLT)\s\d*/;
```
psybers, line 15: Should this be anchored to the start of the string?

Should it be `\d+` instead of 0 or more digits?

psybers (changes requested):

src/app-gocardless/banks/virgin_nrnbgb22.js
matt-fidd: You're right, fixed

psybers:

psybers:

MatissJanis:
Thanks for working on this! Would you mind updating the PR to use prop spreading for `Fallback`? Here's an example bank provider: https://github.com/actualbudget/actual-server/blob/aa22e6951dd90fc03e086593f001d0a002b6e324/src/app-gocardless/banks/american-express-aesudef1.js#L7

We recently merged some changes to simplify them.

Happy to merge this PR once that's done + the conflict is solved!

matt-fidd:
> Thanks for working on this! Would you mind updating the PR to use prop spreading for `Fallback`? Here's an example bank provider:
> 
> https://github.com/actualbudget/actual-server/blob/aa22e6951dd90fc03e086593f001d0a002b6e324/src/app-gocardless/banks/american-express-aesudef1.js#L7
> 
> We recently merged some changes to simplify them.
> 
> Happy to merge this PR once that's done + the conflict is solved!

All done! Thank you

MatissJanis (approved):
Thanks!

PR #361, opened 2024-05-23
Log response when JSON parsing fails

iffy:
I've received bug reports from Actual users for failures at this line and I'd really like to see the response body people are receiving here.

I have NOT tested this code as I don't have Actual set up.


psybers:
This seems like a reasonable thing to log.

Could you fix the linter error and add a release note (see https://github.com/actualbudget/actual-server/blob/master/upcoming-release-notes/README.md)?

MatissJanis:
👋 Hey @iffy !

+1 to what  @psybers said (if you were waiting for an official comment from maintainer :) ). Happy to get this merged as long as CI passes.

PR #362, opened 2024-05-25
Refactor GoCardless bank code to avoid duplication

psybers:
The custom bank code typically copies/pastes an existing bank and then modifies it.  So there is a lot of duplicated code.  This is a first pass at refactoring to reuse the fallback code from `integration-bank.js` and avoid duplication.

NOTE: I do not have GoCardless and can not test for breaking changes.  The existing test suite still works but is quite lacking in coverage of these files.

### DIFF
--- a/src/app-gocardless/banks/american-express-aesudef1.js
+++ b/src/app-gocardless/banks/american-express-aesudef1.js
@@ -1,4 +1,6 @@
-import { amountToInteger, sortByBookingDateOrValueDate } from '../utils.js';
+import Fallback from './integration-bank.js';
+
+import { amountToInteger } from '../utils.js';
 
 /** @type {import('./bank.interface.js').IBank} */
 export default {
@@ -30,7 +32,7 @@ export default {
   },
 
   sortTransactions(transactions = []) {
-    return sortByBookingDateOrValueDate(transactions);
+    return Fallback.sortTransactions(transactions);
   },
 
   /**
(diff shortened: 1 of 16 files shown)


matt-fidd:
I'm looking at adding quite a lot of test coverage to `actual-server` if that's something that would be welcome?

psybers:
> I'm looking at adding quite a lot of test coverage to `actual-server` if that's something that would be welcome?

More tests are always welcome!

src/app-gocardless/banks/american-express-aesudef1.js
```diff
@@ -30,7 +32,7 @@ export default {
   },
 
   sortTransactions(transactions = []) {
-    return sortByBookingDateOrValueDate(transactions);
+    return Fallback.sortTransactions(transactions);
   },
```
MatissJanis, line 36: **💬 suggestion:** ‏might be easier to remove all these defaults entirely and then just use the object spready syntax. WDYT?

```
export default {
  ...Fallback,
  
   sortTransactions() {
       // custom logic
   },
}
```

psybers: Could definitely do that.

And then if a new method has to be added down the road, everything just grabs it by default from the Fallback and the banks do not need updated.

MatissJanis: Yup. That's another benefit (that I definitely thought of before 😄).

MatissJanis (approved):
Thanks

PR #372, opened 2024-06-10
Add bank handler for `NATIONWIDE_NAIAGB21` (Nationwide)

matt-fidd:
Nationwide returns pending transactions with a date representing the latest a transaction could be booked. This stops actual's deduplication logic from working as it only checks 7 days ahead/behind and the transactionID from Nationwide changes when a transaction is booked.

There is also a seperate issue where they sometimes return invalid transactionIds that can change in the future and with the merge of actualbudget/actual#2770 prevents deduplication with the original transaction.

I've reported this to gocardless, they've taken it to nationwide but unfortunately I received this back this morning.
image

An image below showing the issue caused by the transaction ID (these are the same transaction)
image

This has been discussed in Discord <#1148889147801604096> where it was suggested I upstream this patch which I've been using in my own fork for a while.

### DIFF
--- a/src/app-gocardless/bank-factory.js
+++ b/src/app-gocardless/bank-factory.js
@@ -8,6 +8,7 @@ import IngIngddeff from './banks/ing-ingddeff.js';
 import IngPlIngbplpw from './banks/ing-pl-ingbplpw.js';
 import IntegrationBank from './banks/integration-bank.js';
 import MbankRetailBrexplpw from './banks/mbank-retail-brexplpw.js';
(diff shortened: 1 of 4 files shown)


matt-fidd:
I'm having some issues with deduplication that I've not had a chance to look into yet so I'll keep this in WIP until I can verify if this change is the culprit

matt-fidd:
This change isn't the culprit, I've raised actualbudget/actual#2877 for discussion. Opening this back up for review

MatissJanis (approved):
Thanks!

PR #373, opened 2024-06-16
Update docker-release.yml to tag images with all semver levels

jhollowe:
Add tags for all semver levels (major, major.minor, and major.minor.patch) to the release docker images. This allows users to specify a higher level tag when pulling and updating images; users can choose something more stable than `latest` without having to lock to a specific M.m.p


jhollowe:
not stale, waiting on review.

youngcw:
Would there be a use case to lock to, say, version 23.x.x when 23.12.0 has just as much in common with 24.1.0 as it does with 23.11.0?  Since the version scheme is year.month.patch

jhollowe:
lol, It had not clicked with me that it is `year.month.patch`; I guess this isn't really needed.

I'll leave this PR around if maintainers want to merge it, and if not just let the bot close as stale.

twk3:
@jhollowe the `type=semver,pattern={{major}}.{{minor}}` is likely still worth adding if you want to update this PR.  It was lost during a previous change even though the comment still indicates we have it.

jhollowe:
This is ready to be reviewed and merged

PR #381, opened 2024-06-26
Openid

lelemm:
To work in conjunction with [#2945](https://github.com/actualbudget/actual/pull/2945)
Following the old PR #219


src/sql/account/migrations/20000000_old_schema.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE auth
```
MatissJanis, line 1: What are all these migration files in `sql/account/migrations/*`? AFAIK they are unused.

lelemm: I'm not sure. I thought that was already on master. Looks like it was created on the old PR from that I based this on

lelemm: fixed

src/accounts/password.js
```diff
@@ -0,0 +1,62 @@
...
+export function bootstrapPassword(password) {
+  if (!password || password === null || password === '') {
+    return { error: 'invalid-password' };
+  }
+
+  getAccountDb().mutate('DELETE FROM auth WHERE method = ?', ['password']);
+
+  // Hash the password. There's really not a strong need for this
+  // since this is a self-hosted instance owned by the user.
+  // However, just in case we do it.
```
MatissJanis, line 17: **💬 suggestion:** ‏hashing passwords is security 101. Even for local set-ups that should be done. Lets remove the comment.
```suggestion
```

src/accounts/password.js
```diff
@@ -0,0 +1,62 @@
...
+      'SELECT token FROM sessions WHERE user_name = ?',
+      [''],
+    );
+
+    clearExpiredSessions();
+
+    return row.token;
+  } else {
+    return null;
+  }
```
MatissJanis, line 47:
```suggestion
  }
  return null;
```

src/app-sync.js
```diff
@@ -7,12 +7,12 @@ import {
   requestLoggerMiddleware,
   validateUserMiddleware,
 } from './util/middlewares.js';
-import getAccountDb from './account-db.js';
 import { getPathForUserFile, getPathForGroupFile } from './util/paths.js';
 
 import * as simpleSync from './sync-simple.js';
 
 import { SyncProtoBuf } from '@actual-app/crdt';
+import getAccountDb, { isAdmin } from './account-db.js';
```
MatissJanis, line 15: **🔨 warning:** ‏I believe most of the endpoints in this file should be moved behind a permission check. For example: I should not be able to `/sync` a file that I do not have permissions to.

Ideally we should do this in a middleware so it could be bulk-applied to all the relevant endpoints.

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+
+export const TOKEN_EXPIRATION_NEVER = -1;
+
+const sendErrorResponse = (res, status, reason, details) => {
+  res.status(status).send({
+    status: 'error',
+    reason,
+    details,
+  });
+};
```
MatissJanis, line 24: **💭 thought:** ‏personally - I'm not really a fan of this abstraction. It makes the code less readable.

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+
+const getUserByUsername = (userName) => {
+  return (
+    getAccountDb().first('SELECT id FROM users WHERE user_name = ?', [
+      userName,
+    ]) || {}
+  );
+};
```
MatissJanis, line 32: **💬 suggestion:** ‏It feels like a lot of the user and other table SQL queries could be extracted to an `user-service.js` file with readable names. See the example of `secret-service.js` we already have.

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+      'role-does-not-exists',
+      'Selected role does not exists',
+    );
+    return false;
+  }
+
+  return true;
+};
+
+app.get('/masterCreated/', (req, res) => {
```
MatissJanis, line 89: dash-case please
```suggestion
app.get('/owner-created/', (req, res) => {
```

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+  const { cnt } =
```
MatissJanis, line 90: **💬 suggestion:** ‏variable naming could be improved to be more readable

MatissJanis (changes requested):
This is pt1 of my review. Github crashed for me and deleted half of my comments.. So I'll continue the review now and try to re-add them in smaller chunks.

Sorry for being so nitpicky about things here. Getting the `actual-server` part correct is very important. This will get extra scrutiny as compared to the UI implementation in `actual`.

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+     FROM users
+     WHERE users.user_name <> '' and users.master = 1`,
+    ) || {};
+
+  res.json(cnt > 0);
+});
+
+app.get('/users/', (req, res) => {
+  const session = validateUser(req, res);
+  if (!session) return;
```
MatissJanis, line 102: **💭 thought:** ‏this is a good example of why I don't like moving HTTP return statements away from the endpoints: does `validateUser` return something to the customer making this HTTP call? Just by looking at the fn name - I have no idea. I need to dig into the implementation to figure this out.

Would be more readable to do something like..

```suggestion
  const session = validateUser(req, res);
  if (!session) return res.status(401).json({ error: 'unauthenticated' });
```

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+    getAccountDb().first(
+      `SELECT count(*) as cnt
+     FROM users
+     WHERE users.user_name <> '' and users.master = 1`,
+    ) || {};
+
+  res.json(cnt > 0);
+});
+
+app.get('/users/', (req, res) => {
```
MatissJanis, line 100: **❓ question:** ‏should non-admins be able to see the full user list?

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+});
+
+app.post('/users', async (req, res) => {
+  const user = await getAdminSessionFromRequest(req, res);
+  if (!user) return;
+
+  const newUser = req.body;
+  const { id: userIdInDb } = getUserByUsername(newUser.userName);
+
+  if (!validateUserInput(res, newUser)) return;
```
MatissJanis, line 128: **💭 thought:** ‏same problem here: is `validateUserInput` returning some HTTP response? Is this just exiting without returning anything to the end-customer? I have no idea without digging though the implementation of `validateUserInput`

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+  const user = await getAdminSessionFromRequest(req, res);
+  if (!user) return;
+
+  const userToUpdate = req.body;
+  const { id: userIdInDb } =
+    getAccountDb().first('SELECT id FROM users WHERE id = ?', [
+      userToUpdate.id,
+    ]) || {};
+
+  if (!validateUserInput(res, userToUpdate)) return;
```
MatissJanis, line 166: **💬 suggestion:** ‏I think validation should be moved up to line 159 (before querying the DB).

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+
+  getAccountDb().mutate('UPDATE user_roles SET role_id = ? WHERE user_id = ?', [
+    userToUpdate.role,
+    userIdInDb,
+  ]);
+
+  res.status(200).send({ status: 'ok', data: { id: userIdInDb } });
+});
+
+app.post('/users/delete-all', async (req, res) => {
```
MatissJanis, line 193: **❓ question:** is user deletion exposed in the UI? I did not notice it.

Also: this should be a `app.delete('/users', ..`

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+  const user = await getAdminSessionFromRequest(req, res);
+  if (!user) return;
+
+  const ids = req.body.ids;
+  let totalDeleted = 0;
+  ids.forEach((item) => {
+    const { id: masterId } =
+      getAccountDb().first('SELECT id FROM users WHERE master = 1') || {};
+
+    if (item === masterId) return;
```
MatissJanis, line 203: **🔨 warning:** ‏missing HTTP response (400)

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+  if (!session) return;
+
+  if (!checkFilePermission(userAccess.fileId, session.user_id, res)) return;
+
+  if (!userAccess.userId) {
+    sendErrorResponse(res, 400, 'user-cant-be-empty', 'User cannot be empty');
+    return;
+  }
+
+  const { cnt } =
```
MatissJanis, line 290: **💬 suggestion:** ‏variable naming could be improved

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+
+  getAccountDb().mutate(
+    'INSERT INTO user_access (user_id, file_id) VALUES (?, ?)',
+    [userAccess.userId, userAccess.fileId],
+  );
+
+  res.status(200).send({ status: 'ok', data: {} });
+});
+
+app.post('/access/delete-all', (req, res) => {
```
MatissJanis, line 314: **💬 suggestion:** ‏
```suggestion
app.delete('/access', (req, res) => {
```

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+
+  const ids = req.body.ids;
+  let totalDeleted = 0;
+  ids.forEach((item) => {
+    const accessDeleted = getAccountDb().mutate(
+      'DELETE FROM user_access WHERE user_id = ?',
+      [item],
+    ).changes;
+    totalDeleted += accessDeleted;
+  });
```
MatissJanis, line 329: **💬 suggestion:** ‏you can do a batch delete here

```
DELETE FROM user_access WHERE user_id IN ?
```

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+  if (ids.length === totalDeleted) {
+    res
+      .status(200)
+      .send({ status: 'ok', data: { someDeletionsFailed: false } });
+  } else {
+    sendErrorResponse(res, 400, 'not-all-deleted', '');
+  }
+});
+
+app.get('/access', (req, res) => {
```
MatissJanis, line 227: **💬 suggestion:** ‏(applicable to the other endpoints too)
```suggestion
app.get('/users/access', (req, res) => {
```

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+  if (ids.length === totalDeleted) {
+    res
+      .status(200)
+      .send({ status: 'ok', data: { someDeletionsFailed: false } });
+  } else {
+    sendErrorResponse(res, 400, 'not-all-deleted', '');
+  }
+});
+
+app.get('/access/users', async (req, res) => {
```
MatissJanis, line 340: **❓ question:** ‏how is this different from the existing GET /access endpoint? Could both of them be unified?

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+      FROM users
+      LEFT JOIN user_access ON user_access.file_id = ? and user_access.user_id = users.id
+      LEFT JOIN files ON files.id = ? and files.owner = users.id
+      WHERE users.enabled = 1 AND users.user_name <> ''`,
+    [fileId, fileId],
+  );
+  res.json(users);
+});
+
+app.post('/access/get-bulk', (req, res) => {
```
MatissJanis, line 360: **❓ question:** ‏what does this do? Is it used?

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+       WHERE files.id = ?`,
+      [fileId, fileId],
+    );
+    accessMap.set(fileId, userAccess);
+  });
+
+  res.status(200).send({ status: 'ok', data: Array.from(accessMap.entries()) });
+});
+
+app.get('/access/check-access', (req, res) => {
```
MatissJanis, line 383: **❓ question:** ‏do we need this endpoint? Could we instead use the response of the initial sync for this?

i.e. - one of the endpoints returns a list of files for the current user. Then the user clicks on the file to access it. This triggers the download file endpoint. Can we use the "download file endpoint" to check access instead of having a dedicated endpoint for this same operation?

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+  const { method } =
+    getAccountDb().first(
+      `SELECT method from auth
+        where active = 1`,
+    ) || {};
+
+  res.json({ method });
+});
+
+app.get('/multiuser', (req, res) => {
```
MatissJanis, line 487: **💭 thought:** ‏I wonder if we could return the state of multi-user as part of the `bootstrap` endpoint. Or perhaps one of the auth endpoints.

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+        [fileId],
+      ) || {};
+
+    res.json(owner);
+  }
+
+  return null;
+});
+
+app.get('/auth-mode', (req, res) => {
```
MatissJanis, line 477: **❓ question:** ‏what does this do? Why does it return only a single auth method (there can be multiple enabled I think)?

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+
+  getAccountDb().mutate('UPDATE files SET owner = ? WHERE id = ?', [
+    newUserOwner.newUserId,
+    newUserOwner.fileId,
+  ]);
+
+  res.status(200).send({ status: 'ok', data: {} });
+});
+
+app.get('/file/owner', async (req, res) => {
```
MatissJanis, line 441: **❓ question:** ‏is this used to get the owner of a specific file? Could we attach it to the endpoint that serves the list of available files?

src/account-db.js
```diff
@@ -38,31 +51,123 @@ export function getLoginMethod(req) {
...
 }
 
-export function bootstrap(password) {
-  if (password === undefined || password === '') {
-    return { error: 'invalid-password' };
+export async function bootstrap(loginSettings) {
+  const passEnabled = Object.prototype.hasOwnProperty.call(
+    loginSettings,
+    'password',
+  );
```
MatissJanis, line 62:
```suggestion
  const passEnabled = 'password' in loginSettings;
```

src/account-db.js
```diff
@@ -38,31 +51,123 @@ export function getLoginMethod(req) {
...
+  const openIdEnabled = Object.prototype.hasOwnProperty.call(
+    loginSettings,
+    'openid',
+  );
+
+  const { cnt } =
```
MatissJanis, line 68: variable naming

MatissJanis (changes requested):
pt2 (final part)

migrations/1719409568000-multiuser.js
```diff
@@ -0,0 +1,53 @@
...
+    DELETE FROM sessions;
+
+    ALTER TABLE sessions
+        ADD COLUMN expires_at INTEGER;
+
+    ALTER TABLE sessions
+        ADD user_id TEXT;
+
+    ALTER TABLE sessions
+        ADD auth_method TEXT;
```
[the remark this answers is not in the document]
lelemm: I tried that initially, but belive or not, sqlite does not support multiple column alter:
https://stackoverflow.com/questions/6172815/sqlite-alter-table-add-multiple-columns-in-a-single-statement

src/load-config.js
```diff
@@ -72,6 +71,9 @@ let defaultConfig = {
     fileSizeLimitMB: 20,
   },
   projectRoot,
+  password: '',
```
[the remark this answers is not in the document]
lelemm: not sure if it was a test that i forgot to remove or some legacy code from the old PR. anyway it's not even in use. removing it

src/util/middlewares.js
```diff
@@ -20,10 +20,12 @@ async function errorMiddleware(err, req, res, _next) {
...
  */
 const validateUserMiddleware = async (req, res, next) => {
-  let user = await validateUser(req, res);
-  if (!user) {
+  let session = await validateUser(req, res);
+  if (!session) {
     return;
   }
+
+  req.userSession = session;
```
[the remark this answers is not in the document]
lelemm: I tried to implement a simple solution. there is https://github.com/expressjs/session express-session. It looks overkill. Never used it but if you understand that this is the direction to go I can learn how to use and go from there.

migrations/1719409568000-multiuser.js
```diff
@@ -0,0 +1,53 @@
+import getAccountDb from '../src/account-db.js';
+
+export const up = async function () {
+  await getAccountDb().exec(
+    `CREATE TABLE users
+        (id TEXT PRIMARY KEY,
+        user_name TEXT, 
+        display_name TEXT,
+        enabled INTEGER,
+        master INTEGER);
```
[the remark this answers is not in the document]
lelemm: basically an admin is any user defined as admin role.
an master/owner (I'm renaming all to owner) is the owner of the server. this owner user cannot be deleted for safety reasons. It can be changed using the interface, but never deleted so there will be always a user with admin rights

migrations/1719409568000-multiuser.js
```diff
@@ -0,0 +1,53 @@
...
+
+    CREATE TABLE roles
+      (id TEXT PRIMARY KEY,
+      permissions TEXT,
+      name TEXT);
+
+    INSERT INTO roles VALUES ('213733c1-5645-46ad-8784-a7b20b400f93', 'ADMINISTRATOR','Admin');
+    INSERT INTO roles VALUES ('e87fa1f1-ac8c-4913-b1b5-1096bdb1eacc', '','Basic');
```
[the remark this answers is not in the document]
lelemm: We can simplify. I left this way as a open to custom permissions for the future.

src/accounts/password.js
lelemm: agreed. this probably comes from the old PR

src/util/middlewares.js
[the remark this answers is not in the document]
lelemm: So, I investigated express-session and it basically is an overhead (for now) for what we are trying to achieve.
The way we add user data to session is basically the same as we are doing:
image
this is the doc from express-session on how to add custom data into session.

It makes sense when using typescript, but since server is just javascript. I suggest to keep the current implementation and when we move the server from js to ts we implement a more robust typechecking and session handler as express-session. WDYT @MatissJanis ?

migrations/1719409568000-multiuser.js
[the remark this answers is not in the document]
lelemm: introducing a random value will make it complex to validate on client side, adding one more route just for that

migrations/1719409568000-multiuser.js
```diff
@@ -0,0 +1,53 @@
...
+  );
+};
+
+export const down = async function () {
+  await getAccountDb().exec(
+    `
+      DROP TABLE users;
+      DROP TABLE roles;
+      DROP TABLE user_roles;
+      `,
```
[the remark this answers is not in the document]
lelemm: done.
sqlite does not have drop column, so i had to recreate the table without the column

src/app-admin.js
lelemm: not in use. removed

src/app-admin.js
lelemm: i will implement this

src/app-admin.js
lelemm: nnot used. removed

src/app-account.js
```diff
@@ -79,6 +107,72 @@ app.post('/login', (req, res) => {
...
+
+  if (error) {
+    res.status(400).send({ status: 'error', reason: error });
+    return;
+  } else {
+    res.send({ status: 'ok' });
+  }
+});
+
+app.get('/openid-config', async (req, res) => {
```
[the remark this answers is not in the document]
lelemm: done

src/app-account.js
```diff
@@ -79,6 +107,72 @@ app.post('/login', (req, res) => {
...
+    return;
+  }
+
+  res.send({ openId: JSON.parse(auth.extra_data) });
+});
+
+app.get('/login-openid/cb', async (req, res) => {
+  let { error, url } = await loginWithOpenIdFinalize(req.query);
+  if (error) {
+    res.send({ error });
```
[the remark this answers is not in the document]
lelemm: fixed

src/app-admin.js
lelemm: improved

src/app-admin.js
lelemm: solved

src/app-admin.js
lelemm: done

src/app-admin.js
lelemm: moved

src/app-admin.js
lelemm: yes, it is exposed on USER DIRECTORY page.

changed to delete verb

src/app-admin.js
lelemm: this is important to return not only the id, but the username / displayname so we can show it in the file list. it can be baked into the file list. i will do it

src/app-admin.js
lelemm: if they have a budget, they need to know the user list to give access. since the route needs a session, it's not totally public, only users from that instance can get user list

src/app-admin.js
```diff
@@ -0,0 +1,444 @@
...
+
+let app = express();
+app.use(express.json());
+app.use(express.urlencoded({ extended: true }));
+app.use(errorMiddleware);
+app.use(requestLoggerMiddleware);
+
+export { app as handlers };
+
+app.get('/ownerCreated/', (req, res) => {
```
MatissJanis, line 21: **💬 suggestion:** ‏
```suggestion
app.get('/owner-created/', (req, res) => {
```

src/app-admin.js
MatissJanis: bump

MatissJanis (commented):
Overall looks very good. Just a couple of smaller comments. I'm really hopeful we could merge this after the mergefreeze is lifted.

Apologies again about the review taking so long.

src/config-types.ts
```diff
@@ -2,7 +2,7 @@ import { ServerOptions } from 'https';
 
 export interface Config {
   mode: 'test' | 'development';
-  loginMethod: 'password' | 'header';
+  loginMethod?: 'password' | 'header' | 'openid' | null;
```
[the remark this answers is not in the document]
lelemm: reverted it. not sure why is that

src/load-config.js
```diff
@@ -48,7 +48,6 @@ if (process.env.ACTUAL_CONFIG_PATH) {
 
 /** @type {Omit<import('./config-types.js').Config, 'mode' | 'dataDir' | 'serverFiles' | 'userFiles'>} */
 let defaultConfig = {
-  loginMethod: 'password',
```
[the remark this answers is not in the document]
lelemm: got it back

src/app-admin.js
lelemm: changed

migrations/1718889148000-openid.js
```diff
@@ -0,0 +1,35 @@
...
+  );
+};
+
+export const down = async function () {
+  await getAccountDb().exec(
+    `
+      CREATE TABLE auth_new
+        (password TEXT);
+      INSERT INTO auth_new (password)
+        SELECT extra_data FROM auth;
```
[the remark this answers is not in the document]
lelemm: fixed

migrations/1719409568000-multiuser.js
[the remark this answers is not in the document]
lelemm: done

src/accounts/openid.js
```diff
@@ -0,0 +1,224 @@
+import getAccountDb, { clearExpiredSessions } from '../account-db.js';
+import * as uuid from 'uuid';
+import { generators, Issuer } from 'openid-client';
+import { TOKEN_EXPIRATION_NEVER } from '../app-admin.js';
+import finalConfig from '../load-config.js';
+
+export async function bootstrapOpenId(config) {
+  if (!Object.prototype.hasOwnProperty.call(config, 'issuer')) {
```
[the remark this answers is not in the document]
lelemm: done

src/accounts/openid.js
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  let accountDb = getAccountDb();
+  let config = accountDb.first('SELECT extra_data FROM auth WHERE method = ?', [
+    'openid',
+  ]);
+  if (!config) {
+    return { error: 'openid-not-configured' };
+  }
+  config = JSON.parse(config['extra_data']);
```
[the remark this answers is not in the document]
lelemm: done

src/accounts/openid.js
```diff
@@ -0,0 +1,224 @@
...
+      userId = uuid.v4();
+      accountDb.mutate(
+        'INSERT INTO users (id, user_name, display_name, enabled, master) VALUES (?, ?, ?, 1, 1)',
+        [userId, identity, userInfo.name ?? userInfo.email ?? identity],
+      );
+
+      accountDb.mutate(
+        'INSERT INTO user_roles (user_id, role_id) VALUES (?, ?)',
+        [userId, '213733c1-5645-46ad-8784-a7b20b400f93'],
+      );
```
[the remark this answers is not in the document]
lelemm: removed user_roles, added field ROLE in the users table. no more IDS, just the string ADMIN and ROLE

src/app-account.js
```diff
@@ -79,6 +107,72 @@ app.post('/login', (req, res) => {
   res.send({ status: 'ok', data: { token } });
 });
 
+app.post('/enable-openid', async (req, res) => {
+  const session = await getAdminSessionFromRequest(req, res);
+  if (!session) return;
```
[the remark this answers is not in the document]
lelemm: refactored that

src/app-account.js
```diff
@@ -79,6 +107,72 @@ app.post('/login', (req, res) => {
...
+
+  let { error } = (await enableOpenID(req.body)) || {};
+
+  if (error) {
+    res.status(400).send({ status: 'error', reason: error });
+    return;
+  } else {
+    res.send({ status: 'ok' });
+  }
```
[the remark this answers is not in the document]
lelemm: done

src/load-config.js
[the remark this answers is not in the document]
lelemm: removed

src/util/middlewares.js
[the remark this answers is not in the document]
lelemm: changed to res.locals

src/app-account.js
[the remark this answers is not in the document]
lelemm: done

src/app-account.js
```diff
@@ -79,6 +107,72 @@ app.post('/login', (req, res) => {
...
 });
 
+app.post('/enable-openid', async (req, res) => {
+  const session = await getAdminSessionFromRequest(req, res);
+  if (!session) return;
+
+  let { error } = (await enableOpenID(req.body)) || {};
+
+  if (error) {
+    res.status(400).send({ status: 'error', reason: error });
```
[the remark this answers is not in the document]
lelemm: done

src/app-account.js
```diff
@@ -79,6 +107,72 @@ app.post('/login', (req, res) => {
...
+    return;
+  } else {
+    res.send({ status: 'ok' });
+  }
+});
+
+app.post('/enable-password', async (req, res) => {
```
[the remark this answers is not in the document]
lelemm: refactored this

src/app-account.js
```diff
@@ -79,6 +107,72 @@ app.post('/login', (req, res) => {
...
+app.get('/openid-config', async (req, res) => {
+  const { cnt } =
+    getAccountDb().first(
+      `SELECT count(*) as cnt
+   FROM users
+   WHERE users.user_name <> '' and users.master = 1`,
+    ) || {};
+
+  if (cnt > 0) {
+    res.send({});
```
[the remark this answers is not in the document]
lelemm: using 500 when needed now

src/app-secrets.js
```diff
@@ -10,12 +11,28 @@ const app = express();
...
+  const { method } =
+    getAccountDb().first('SELECT method FROM auth WHERE active = 1') || {};
+
   const { name, value } = req.body;
 
+  if (method === 'openid') {
+    let canSaveSecrets = isAdmin(req.userSession);
+
+    if (!canSaveSecrets) {
+      res.status(400).send({
```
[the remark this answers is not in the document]
lelemm: done

src/app.js
```diff
@@ -83,5 +88,8 @@ export default async function run() {
   } else {
     app.listen(config.port, config.hostname);
   }
+
+  if (!(await toogleAuthentication())) exit(-1);
```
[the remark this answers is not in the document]
lelemm: added logs to toggleAuthentication()

src/util/validate-user.js
```diff
@@ -26,6 +27,19 @@ export default function validateUser(req, res) {
     return null;
   }
 
+  if (
+    session.expires_at !== TOKEN_EXPIRATION_NEVER &&
+    session.expires_at * 1000 <= Date.now()
+  ) {
+    res.status(403);
```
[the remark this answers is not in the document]
lelemm: done

src/app-admin.js
lelemm: removed it in favor of use the modified /list-user-files

src/app-admin.js
lelemm: moved it to "/needs-bootstrap"

src/app-admin.js
```diff
@@ -0,0 +1,491 @@
...
+      `SELECT files.owner
+     FROM files
+     WHERE files.id = ?`,
+      [fileId],
+    ) || {};
+
+  res.json({ granted: owner === session.user_id });
+});
+
+app.post('/access/transfer-ownership/', (req, res) => {
```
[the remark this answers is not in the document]
lelemm: yes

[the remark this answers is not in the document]
lelemm: this is used on user access page under the "..." menu. a admin or file owner can transfer the ownership to someone else in case this user is removed.

lelemm:
Closing this PR in favor of a new one because this one is very slow with all the comments from code rabbit over time

lelemm:
closed

PR #384, opened 2024-07-02
Sync multiple accounts in a single SimpleFIN API call.

psybers:
This is the server implementation that supports SimpleFIN sync'ing one account (as previously supported) *or* several accounts in a single call.  The client side would need to be updated to utilize this.

Feedback welcome.


youngcw (approved):
This all makes sense to me.

endigma:
What does this need to be merge-ready? I can't really use Actual until this gets fixed as every refresh cycle locks up my SimpleFIN access.

psybers:
Honestly, the back-end code could merge at any time.  But the front end won't use the new functionality.

psybers:
Ok so one thing I failed to think through on the first attempt was that maybe some accounts sync and some fail.  So if we try all, what do we do in such a case?  Previously, it would see the first failure and just abandon the entire attempt.

My current thought is we can still try to parse out the other accounts and maybe get a partial update.  So account-specific errors are simply stored while processing and then at the end, if there was one account being sync'd and there is an error I have it respond with that error.  If there were multiple accounts and at least one error, I return the transaction data we found and then set a new `errors` field in the object.  The client will have to know to look for those errors and show them on specific accounts.

psybers:
I really need someone else to test this.

natrlhy:
How can I test?

psybers:
> How can I test?

You have to run this branch of the server and then connect to it with a client.  Then try sync'ing.

tcrasset:
I will not be able to test it as I don't have a simplefin account, nor do I have a bank account that I could sync (I use GoCardless). Though, I'm working on a refactor (https://github.com/actualbudget/actual-server/pull/440) that will make it so that we can easily fake out responses from the SimplefinAPI, so that we don't have to rely on manual testing. Because, in the end, most of the code added in this PR is just transforming the simplefin responses into a different format.

matt-fidd (commented):
Looks good from a code review stance, unfortunately I'm not able to test properly with SF so won't give it a full approval.

One tiny wording nit to make it a bit clearer, not a blocker though.

youngcw:
@psybers Any reason I shouldn't merge this in? (or Matt can if he beats me to it)

psybers:
@youngcw It should be fine to merge this. The behavior of this code should be the same as the existing code, but opens the ability to process multiple accounts in a single call.

The idea is to update the front end next to utilize that new functionality.

matt-fidd (approved):

youngcw:
> Is there a tracking issue for front-end support?

@endigma I don't think so

PR #386, opened 2024-07-03
Prettify GoCardless Bank Integration README

tcrasset:

### DIFF
--- a/src/app-gocardless/README.md
+++ b/src/app-gocardless/README.md
@@ -1,47 +1,164 @@
 # Integration new bank
 
-Find in [doc](https://docs.google.com/spreadsheets/d/1ogpzydzotOltbssrc3IQ8rhBLlIZbQgm5QCiiNJrkyA/edit#gid=489769432) what is id of bank which you want to integrate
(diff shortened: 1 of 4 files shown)


src/app-gocardless/README.md
```diff
@@ -1,47 +1,164 @@
 # Integration new bank
 
-Find in [doc](https://docs.google.com/spreadsheets/d/1ogpzydzotOltbssrc3IQ8rhBLlIZbQgm5QCiiNJrkyA/edit#gid=489769432) what is id of bank which you want to integrate
+If the default bank integration does not work for you, you can integrate a new bank by following these steps.
```
tcrasset, line 3: I tried to get bank sync to work, and it wouldn't (due to my Adblocker, duh).

So I figured that each bank needed a custom integration.

I then went ahead and tried to setup my bank but had a hard time.

It is only after that I tried without Adblocker and it went fine.

This small message would have helped me in trying harder with the default.

MatissJanis (approved):
Thanks!

PR #388, opened 2024-07-06
remove version from docker-compose.yml

### LINKED ISSUE #387: [Bug]: Docker-compose.yml - version is obsolete - warning on command line

### What happened?

When you run `docker compose`, you get a warning that "version" is obsolete. the warning is saying that the line "version" in compose-yml is obsolete per the spec, but the message could be confusing to new users.
https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-and-name-top-level-elements

time="2024-07-06T11:18:23-10:00" level=warning msg="actual-server\\docker-compose.yml: `version` is obsolete"

Docker

Other

### Operating System

Windows 10

wdpk:
version for docker compose files is obsolete, it has no practical functionality, but causes WARN on command line. previously it was used to determine the type of dockerfile, and saying " "version" is obsolete" might be confusing to non-technical users trying to get actual running via docker.

fixes #387

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: '3'
 services:
   actual_server:
     image: docker.io/actualbudget/actual-server:latest


MatissJanis (approved):
Thanks

PR #391, opened 2024-07-08
Update SEB Credit Card adapter to include SEB Card Accounts

craigmdennis:
Relates to #261 
Add SEB Card Accounts institution ID to correctly import purchases as payments instead of deposits.

### Before
Arc 2024-07-08 at 10 56 37@2x

### After
Arc 2024-07-08 at 10 58 40@2x


MatissJanis (approved):
Thanks

PR #396, opened 2024-07-13
Add gocardless support for easybank (Austria)

neuos:
This enhances the parsing of payeeName for the Austrian bank easybank (BAWAATWW) as it is often left blank with the default implementation.


neuos:
Every custom Bank mapper sets the `payeeName`, but it is not specified in the return type of `normalizeTransaction`.
So maybe I fixed a bug, or I didn't understand the model quite right, I am looking forward to an explanation :)

MatissJanis (approved):

PR #398, opened 2024-07-13
[WIP] Add a command to reset the password on PikaPods

noamraph:
This adds a script which allows to easily reset the password on PikaPods.

This addresses https://github.com/actualbudget/actual/issues/3052.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "types": "tsc --noEmit --incremental",
     "verify": "yarn -s lint && yarn types",
     "reset-password": "node src/scripts/reset-password.js",
+    "reset-password-pikapods": "node src/scripts/reset-password-pikapods.js",
     "health-check": "node src/scripts/health-check.js"
   },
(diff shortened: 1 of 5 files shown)


tcrasset:
Thank you for this, I am not a maintainer, but I have some comments to make. Feel free to disregard them.
Wouldn't it be easier just to expose an environment variable that takes in the new password, and the server modifies it upon restart when it sees that the environment variable is set?


The current proposed solution is quite involved, needing to change some parameters to PikaPods, download a database, modifying the database, uploading the database again, disabling SFTP access again.

There are many steps involved, which might corrupt the database. Moreover, it's harder/impossible to test.

noamraph:
@tcrasset Nice idea! I agree that adding an environment variable would be simpler. I'm closing this, hopefully I'll have time to create a PR to use such an envvar, say ACTUAL_RESET_PASSWORD_TO

Issue #392, opened 2024-07-09
[Bug]: express-rate-limit validation error when using with reverse proxy


ankel:
### Verified issue does not already exist?

- [X] I have searched and found no existing issue

### What happened?

Reproduce steps: run `docker compose up` with the following configurations

compose.yaml
```
services:
  traefik:
    image: traefik:latest
    restart: unless-stopped
    ports:
      - "80:80"
    volumes:
      - "./traefik.yaml:/etc/traefik/traefik.yaml"
      - "./traefik/data:/data"
      - "/var/run/docker.sock:/var/run/docker.sock"

  actual-server:
    image: actualbudget/actual-server:latest-alpine
    restart: unless-stopped
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.actual-server.rule=PathPrefix(`/actual-budget`)"
      - "traefik.http.routers.actual-server.entrypoints=web"
      - "traefik.http.services.actual-server.loadbalancer.server.port=5006"
    volumes:
      - ./actual-data:/data
```

traefik.yaml
```
entryPoints:
  web:
    address: ":80"

providers:
  docker: {}
```

### What error did you receive?

Once it's running, `curl localhost:80/actual-budget` will print the following error in the log:

```
ValidationError: The 'X-Forwarded-For' header is set but the Express 'trust proxy' setting is false (default). This could indicate a misconfiguration which would prevent express-rate-limit from accurately identifying users. See https://express-rate-limit.github.io/ERR_ERL_UNEXPECTED_X_FORWARDED_FOR/ for more information.
    at _Validations.<anonymous> (file:///app/node_modules/express-rate-limit/dist/index.mjs:154:15)
    at _Validations.wrap (file:///app/node_modules/express-rate-limit/dist/index.mjs:287:18)
    at _Validations.xForwardedForHeader (file:///app/node_modules/express-rate-limit/dist/index.mjs:152:10)
    at Object.keyGenerator (file:///app/node_modules/express-rate-limit/dist/index.mjs:516:19)
    at file:///app/node_modules/express-rate-limit/dist/index.mjs:569:32
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async file:///app/node_modules/express-rate-limit/dist/index.mjs:550:5 {
  code: 'ERR_ERL_UNEXPECTED_X_FORWARDED_FOR',
[...]

PR #379, opened 2024-06-19
fix client IP checked against trusted proxies

tuetenk0pp:
fixes #371

Issue #371, opened 2024-06-09
[Bug]: ACTUAL_TRUSTED_PROXIES has not desired effect


tuetenk0pp:
### Verified issue does not already exist?

- [X] I have searched and found no existing issue

### What happened?

I set up actual budget to allow for header authentication via authentik. So I added the group attributes in authentik:

```yaml
additionalHeaders:
  X-Actual-Password: ***
```

I also added the following environment variables to my docker setup:

```env
ACTUAL_LOGIN_METHOD=header
ACTUAL_TRUSTED_PROXIES=<proxy_ip>
```

This is how I configured Caddy:

```Caddyfile

(header) {
        header {
                Strict-Transport-Security "max-age=31536000; includeSubdomains"
                X-XSS-Protection "1; mode=block"
                X-Content-Type-Options "nosniff"
                Referrer-Policy "same-origin"
                X-Frame-Options "ALLOW-FROM *.<domain.tld>"
                -Server
                Content-Security-Policy "frame-ancestors <domain.tld> *.<domain.tld>"
                Permissions-Policy "geolocation=(self <domain.tld> *.<domain.tld>), microphone=()"
        }
}

(errors) {
        handle_errors {
                rewrite * /500.html
                file_server
        }
}
https://budget.domain.tld {
        # use header snippet
        import header

        # always forward outpost path to actual outpost
        reverse_proxy /outpost.goauthentik.io/* http://<authentik_ip>:9000

        # forward authentication to outpost
        forward_auth http://<authentik_ip>:9000 {
                uri /outpost.goauthentik.io/auth/caddy

                # capitalization of the headers is important, otherwise they will be empty
                copy_headers X-Authentik-Username X-Authentik-Groups X-Authentik-Email X-Authentik-Name X-Authentik-Uid X-Authentik-Jwt X-Authentik-Meta-Jwks X-Authentik-Meta-Outpost X-Authentik-Meta-Provider X-Authentik-Meta-App X-Authentik-Meta-Version X-Actual-Password

                # optional, in this config trust all private ranges, should probably be set to the outposts IP
                trusted_proxies private_ranges
        }

        
        encode gzip zstd
        reverse_proxy http://<actual_budget_ip>:5006

        # use errors snippet
        import errors
}
```

### What error did you receive?

When I try to login into my actual budget instance, I get this error: 

![grafik](https://github.com/actualbudget/actual-server/assets/55783419/35929ead-7a8c-4241-bb49-8ae162868a6f)

[...]

PR #399, opened 2024-07-15
[WIP] Use express builtin 'trust proxy'

djm2k:
### Changes made
- Added `quote_type = single` to `.editorconfig` (VSCode kept changing them all to double).
- Adjust the `trustedProxies` config option to work with [expressjs](https://expressjs.com/en/guide/behind-proxies.html).
- Remove the `validateAuthHeader` function and check.

### Reasoning
- https://github.com/actualbudget/actual-server/pull/312 introduced `validateAuthHeader` which attempted to re-implement logic provided by expressjs itself through `app.set('trust proxy')` (see docs above)

### Linked
- Fixes https://github.com/actualbudget/actual-server/issues/392
- Related to https://github.com/actualbudget/actual-server/issues/371
- Supersedes https://github.com/actualbudget/actual-server/pull/379

### Ongoing work
- [ ] E2E tests or manual testing to confirm working in different contexts i.e. `trustedProxies` set to:
  - [ ] `1` with Traefik/Caddy
  - [ ] `2` with NGINX and Cloudflare proxies
  - [ ] `10.0.0.0/16,0.233.192.169/32` for manual IP address/subnet CSV
  - [ ] `false` to throw `ERR_ERL_UNEXPECTED_X_FORWARDED_FOR` when a proxy header is detected (current behaviour)
  - [x] `true` to throw `ERR_ERL_PERMISSIVE_TRUST_PROXY` 
- [ ] PR for https://github.com/actualbudget/docs explaining the env var better and linking to the express docs


djm2k:
This still needs to be debugged, I am getting this error now after these changes: (implying `trustedProxies` is coming through as `true` instead of the default array)
```
ValidationError: The Express 'trust proxy' setting is true, which allows anyone to trivially bypass IP-based rate limiting. See https://express-rate-limit.github.io/ERR_ERL_PERMISSIVE_TRUST_PROXY/ for more information.
    at _Validations.<anonymous> (file:///app/node_modules/express-rate-limit/dist/index.mjs:134:15)
    at _Validations.wrap (file:///app/node_modules/express-rate-limit/dist/index.mjs:287:18)
    at _Validations.trustProxy (file:///app/node_modules/express-rate-limit/dist/index.mjs:132:10)
    at Object.keyGenerator (file:///app/node_modules/express-rate-limit/dist/index.mjs:515:19)
    at file:///app/node_modules/express-rate-limit/dist/index.mjs:569:32
    at async file:///app/node_modules/express-rate-limit/dist/index.mjs:550:5 {
  code: 'ERR_ERL_PERMISSIVE_TRUST_PROXY',
  help: 'https://express-rate-limit.github.io/ERR_ERL_PERMISSIVE_TRUST_PROXY/'
}
```

djm2k:
@twk3 I am very confused by the different terms. What you're describing is known to me as "Trusted IP Ranges" or "IP Authentication". Handling this at the express middleware level instead of in the reverse proxy config seems like a mistake which could lead to future admins misconfiguring their setups.

My previous experience with header authentication has been to set a specific header to a secret value, and ensure that header was cleared from user's requests by the reverse proxy before being set. 

This could be handled by `validateAuthHeader` where it would check the header is present and it's value matches the secret key set in the config.

I'd like to setup an E2E test which covers your use-case as well. For me- the whole app is wrapped in Authelia no matter if i am accessing it locally or externally.

Let me know what you think

twk3:
> @twk3 I am very confused by the different terms. What you're describing is known to me as "Trusted IP Ranges" or "IP Authentication". Handling this at the express middleware level instead of in the reverse proxy config seems like a mistake which could lead to future admins misconfiguring their setups.
> 
> ~My previous experience with header authentication has been to set a specific header to a secret value, and ensure that header was cleared from user's requests by the reverse proxy before being set.~
> 
> ~This could be handled by `validateAuthHeader` where it would check the header is present and it's value matches the secret key set in the config.~ Nevermind, looking at #312 again, this is implemented with `x-actual-password`
> 
> I'd like to setup an E2E test which covers your use-case as well. For me- the whole app is wrapped in Authelia no matter if i am accessing it locally or externally.
> 
> Let me know what you think

Yeah, sorry. That was my fault for using trusted proxy as the name. Should have been `allowedAuthCIDRs` or something like that. And it SHOULD only need to be set if someone wants to lock header auth to a specific endpoint.

To fix the naming situation, we probably need to rename the current impl to `allowedAuthCIDRs`, but populate it by default with whatever folks have set in their `trustedProxies` (to maintain support for the config syntax that we released). But primarily start using trustedProxies for what you have here. (we also need to fix the current implementation so it's not needed to be set if folks just want header auth, without restricting it)

dm-canteen:
Still WIP

djm2k:
Still WIP https://github.com/djm2k/actual-server/commits/add-e2e-test/

KiARC:
This *really* needs to be reopened, it's been two months and Actual still cannot deploy behind a proxy. What is actually needed to move this along?

djm2k:
@KiARC 
> This _really_ needs to be reopened, it's been two months and Actual still cannot deploy behind a proxy. What is actually needed to move this along?

twk3 has generously taken the above discussion and worked it into changes in [`twk3/fix-auth-header-trust`](https://github.com/actualbudget/actual-server/tree/twk3/fix-auth-header-trust) relating to https://github.com/actualbudget/actual-server/issues/371

Would you like to test the above and report back in https://github.com/actualbudget/actual-server/issues/392?

Please check the repo's activity in the future: https://github.com/actualbudget/actual-server/activity

I will continue working on my [`add-e2e-test`](https://github.com/djm2k/actual-server/tree/add-e2e-test) branch when i get the chance

PR #401, opened 2024-07-18
Added GoCardless Integration for ING (Romania)

spideraxal:
Hello,

I created a custom adapter for ING Bank (Romania). Due to issues with how the transactionId is populated by the bank, new transactions would not be imported to Actual. 

As I am not primarily a developer, feedback is greatly appreciated!


MatissJanis:
👋 Hi!
> As I am not primarily a developer, feedback is greatly appreciated!

That's totally fine. We all have to start somewhere.

Did you test this change locally to verify that it works to your satisfaction? Generally with custom bank-mappers I trust the authors to do the due diligence as there is very limited possibility for me to test this code on my own (since I'm not a customer of this bank).

spideraxal:
Yep, tested it with my own transactions. I still need to 100% confirm the pending transaction deduplication, which should happen in one or two days. I can update the post here before this gets approved.

Also, @matt-fidd helped with the code, so I know this should work 😄

MatissJanis (approved):
Sweet!

PR #402, opened 2024-07-19
Obfuscate password when login the login header

rodriguestiago0:
Obfuscate password when login the login header


src/app-account.js
```diff
@@ -45,7 +45,19 @@ app.post('/login', (req, res) => {
...
+        obfuscated = headerVal.replace(
+          regex,
+          '$1' + '*'.repeat(headerVal.length - 1) + '$2',
+        );
+      } else if (headerVal.length > 0) {
+        obfuscated = headerVal[0] + '*'.repeat(headerVal.length - 1);
+      } else {
+        obfuscated = 'No password provided.';
+      }
+      console.debug('HEADER VALUE: ' + obfuscated);
```
MatissJanis, line 60: **💭 thought:** ‏quite a lot of code to obfuscate this thing. Could we simplify it to something much more _stupid_? That would reduce the chances of this code having bugs that we hadn't considered.

```suggestion
      const obfuscated = '*'.repeat(headerVal.length) || 'No password provided.';
      console.debug('HEADER VALUE: ' + obfuscated);
```

rodriguestiago0: Just wanted to give some characters to help validate if the received password was valid or not.

But I’m ok with it

MatissJanis:
LGTM - just linter failing

MatissJanis (approved):

PR #404, opened 2024-07-22
Revert "Add bank handler for `NATIONWIDE_NAIAGB21` (Nationwide) (#372)"

matt-fidd:
This reverts #372 

There are reports of this handler causing problems in discord (https://discord.com/channels/937901803608096828/1265017131930751037). Probably best to remove for this release.


MatissJanis (approved):

matt-fidd:
Looking at this again, on second thoughts this needs to stay in to fix the dedupe problems.

A better approach would be to remove the arbitrary 8 day adjustment and instead clamp the date to todays date.

PR #405, opened 2024-07-24
Improve GoCardless adapter for ING (Romania)

spideraxal:
Hello,

Based on my last PR, I observed that the notes of a pending transactions did not get updated once it became booked. That meant, you could not create any rules for automatic categorization. This PR will fix that.

Another issue I am currently battling is the starting balance is not calculated correctly when there are pending transactions: 
- The starting balance somehow subtracts the pending balance as well, so when the transactions are added, they get subtracted twice from the account total (example in the screenshots)
- I emailed GoCardless, they provide the right account balance (like in the banking app) so it's not an issue on their side and they cannot provide any other budget type (like interimBooked) as the bank does not send that data over

Any input how this could be solved? I thought maybe accessing the pendingTransactions array (from within calculateStartingBalance) and adding that amount to the starting balance, but that does not seem possible.

Thank you!
image
image
image
image
(Balance is off by 50,51, exactly the pending amount - at the time of those screenshots)


matt-fidd (approved):
Thanks for this!

PR #410, opened 2024-08-04
Stop server crashing when SimpleFIN is down.

psybers:
When SimpelFIN is down, we get 502 errors.  The JSON tries to parse, and fails.  The last commit simply wrapped that in a reject, but then nothing ever catches that rejected promise.  This catches them and sends a failure response code, and then returns (and avoids the server crashing).

### DIFF
--- a/src/app-simplefin/app-simplefin.js
+++ b/src/app-simplefin/app-simplefin.js
@@ -60,7 +60,7 @@ app.post(
         },
       });
     } catch (e) {
-      serverDown(res);
+      serverDown(e, res);
       return;
     }
   }),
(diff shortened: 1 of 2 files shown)


src/app-simplefin/app-simplefin.js
```diff
@@ -50,14 +50,19 @@ app.post(
...
-        accounts: accounts.accounts,
-      },
-    });
+      res.send({
+        status: 'ok',
+        data: {
+          accounts: accounts.accounts,
+        },
+      });
+    } catch (e) {
```
MatissJanis, line 62: **💬 suggestion:** ‏Could you also add a `console.log` in these catch blocks? So we have a bit more debug information if-need-be.

psybers: Done

MatissJanis (approved):

PR #413, opened 2024-08-08
Return early from SimpleFIN /transactions if the account needs attention

jpetso:
Without this, the subsequent res.send() would throw an error for writing headers when another response has already been sent.


MatissJanis:
I think this makes sense.

cc @psybers would you mind verifying?

psybers:
> I think this makes sense.
> 
> cc @psybers would you mind verifying?

Was this already fixed by #412?

psybers:
Ok now that I am on web I can see the line numbers changed quite a bit, so this particular fix was not yet handled.

psybers (approved):

matt-fidd (approved):
Approved based on @psybers comment

Issue #420, opened 2024-08-14
[Feature Request]: apply separation of concerns to this repo


tcrasset:
### Verified issue does not already exist?

- [X] I have searched and found no existing issue

### What happened?

I have read through the code to see how one could export an OpenApi schema as asked by https://github.com/actualbudget/actual-server/issues/419, but what I've found is that there is a deeper issue going on in this repo.

That is, separating concerns in the app, having code whose purpose is handling the incoming HTTP route, code whose purpose it generating a response, and code whose purpose is accessing the database.

Currently, most of the endpoints have all this in a single function. Example:

https://github.com/actualbudget/actual-server/blob/eec5fbb1cc515cc4a2495cc2c683024e5446fb88/src/app-sync.js#L24

There seem to be some files that are going in a positive direction, e.g.

https://github.com/actualbudget/actual-server/blob/master/src/services/secrets-service.js

but it does not seem to be a majority, as far as I could tell.

I'm not criticizing the work that has already been done, obviously there is a lot of work behind all of it, and it works!

However, going forward to be able to scale we would need to have some separation to make addition of routes, testing, adding an OpenAPI spec, much easier.

I would be glad to give a hand, I have dabbled a bit in JS, but my day job consist of being a backend developer in Python, so I know my way around a clean architecture, I'll just need to apply it to JS, similar to the secrets-service.js shown above.

### Where are you hosting Actual?

None

### Operating System

None

tcrasset:
Example of a blog post of how to apply this to an Express JS app:

https://medium.com/@ben.dev.io/clean-architecture-in-node-js-39c3358d46f3

PR #421, opened 2024-08-15
[Maintenance] improve testing utils and add delete-user-file test

tcrasset:
Adding integration tests is a pre-requisite to https://github.com/actualbudget/actual-server/issues/420, as refactoring might break some stuff without us knowing, and manually doing the actions on the front-end is a pain.

I wanted to create a small PR that shows the value of adding integration tests.
I kept it small on purpose so that I wouldn't have done this for nothing if it gets rejected.

My plan is that this gets accepted, and I/we start adding integration test on all/major endpoints, so that we can more readily refactor the backend to be inline with https://github.com/actualbudget/actual-server/issues/420.


package.json
```diff
@@ -9,6 +9,10 @@
     "lint": "eslint . --max-warnings 0",
     "build": "tsc",
     "test": "NODE_ENV=test NODE_OPTIONS='--experimental-vm-modules --trace-warnings' jest  --coverage",
+    "db-migrate": "NODE_ENV=development node src/run-migrations.js up",
+    "db-downgrade": "NODE_ENV=development node src/run-migrations.js down",
+    "db-test-migrate": "NODE_ENV=test node src/run-migrations.js up",
+    "db-test-downgrade": "NODE_ENV=test node src/run-migrations.js down",
```
tcrasset, line 15: When adding tests I often came across the problem that the teardown was not called correctly, which led to the setup failing upon restarting the test suite.

I've added these pair of scripts so that one can manually clear the test database using `npm run db-test-downgrade` when that happens.

src/app-sync.js
```diff
@@ -13,8 +13,13 @@ import { SyncProtoBuf } from '@actual-app/crdt';
 
 const app = express();
 app.use(errorMiddleware);
+app.use(express.json());
+
```
tcrasset, line 17: The `/delete-user-file` `req.body` was not being parsed as `json` by default, which necessitates the middleware to be added.

I manually tested that the `/sync` endpoint still works by launching the frontend add doing some syncable stuff, and watching the requests/responses.
It would be nice if there were tests for this function, which is my next plan.

package.json
matt-fidd: Could we match the syntax in actual used to group scripts here?

```suggestion
    "db:migrate": "NODE_ENV=development node src/run-migrations.js up",
    "db:downgrade": "NODE_ENV=development node src/run-migrations.js down",
    "db:test-migrate": "NODE_ENV=test node src/run-migrations.js up",
    "db:test-downgrade": "NODE_ENV=test node src/run-migrations.js down",
```

src/app-sync.test.js
```diff
@@ -75,3 +75,65 @@ describe('/download-user-file', () => {
...
+      .send({ fileId });
+
+    expect(res.statusCode).toEqual(200);
+    expect(res.body).toEqual({ status: 'ok' });
+
+    // Verify that the file is marked as deleted
+    const rows = accountDb.all('SELECT deleted FROM files WHERE id = ?', [
+      fileId,
+    ]);
+    expect(rows[0].deleted).toBe(1);
```
matt-fidd, line 137: tiny nit, I think this would read more obviously as `true` if possible

```suggestion
    expect(rows[0].deleted).toBe(true);
```

tcrasset: I had it add `true` at the beginning, but it's actually `1`, not `true`.
I could always add `const true = 1` and use the variable, but that seems kinda wrong. 
It might be a good idea in the future to wrap the database return objects in a custom class, so that this could be typed + transformed into a boolean.

matt-fidd: Ah of course, it's an integer field, sorry!

matt-fidd: We could use `.toBeTruthy()`?

https://jestjs.io/docs/expect#tobetruthy

tcrasset: Proof:

```
 FAIL  src/app-sync.test.js
  ● /delete-user-file › marks the file as deleted

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: 1

      135 |       fileId,
      136 |     ]);
    > 137 |     expect(rows[0].deleted).toBe(true);
          |                             ^
      138 |   });
      139 | });
      140 |

      at Object.toBe (src/app-sync.test.js:137:29)
```

tcrasset:
Took your comments into account. 

Also, during some more integration testing for `/sync`, I noticed that I actually needed the middleware to be `express.raw` for the `/sync` endpoint to work during integration tests. I would wager that this is specific to `jest` or `supertest`, as I had no issues without it when using the frontend. Same for the `delete-user-file` endpoint, worked prefectly using `curl` without specifying the `express.json()` middleware, but during integration testing it complained the body was empty, as specified in one of my previous comments.

Am not knowledgeable enough in javascript testing to know why that might be.

tcrasset:
Follow up PR is here: #423

matt-fidd (approved):

PR #422, opened 2024-08-15
[Maintenance] refactor user validation into middleware

tcrasset:


package.json
```diff
@@ -7,6 +7,7 @@
   "scripts": {
     "start": "node app",
     "lint": "eslint . --max-warnings 0",
+    "lint-fix": "eslint . --fix",
```
matt-fidd, line 10: Could we match the syntax in actual here?

```suggestion
    "lint:fix": "eslint . --fix",
```

src/app-gocardless/app-gocardless.js
```diff
@@ -20,13 +20,7 @@ app.get('/link', function (req, res) {
...
 export { app as handlers };
 app.use(express.json());
-app.use(async (req, res, next) => {
-  let user = await validateUser(req, res);
-  if (!user) {
-    return;
-  }
-  next();
-});
+app.use(errorMiddleware);
```
matt-fidd, line 23: should this also use `validateUserMiddleware`?

matt-fidd (changes requested):
Thanks for this! A couple of minor comments

src/app-gocardless/app-gocardless.js
tcrasset: Oops, imported the wrong middleware

tcrasset:
Should be good now. Thanks for the quick review.

matt-fidd (approved):

PR #423, opened 2024-08-15
add integration tests for the `/sync` endpoint

tcrasset:
~Needs https://github.com/actualbudget/actual-server/pull/421 to be merged first.~

Ready to be reviewed!

PR #425, opened 2024-08-15
add more tests for app-sync.js

tcrasset:
~Needs #423 to be merged first~

### DIFF
--- a/src/app-sync.js
+++ b/src/app-sync.js
@@ -12,7 +12,9 @@ import { SyncProtoBuf } from '@actual-app/crdt';
 
 const app = express();
 app.use(errorMiddleware);
+app.use(express.json());
 app.use(express.raw({ type: 'application/actual-sync' }));
+app.use(express.raw({ type: 'application/encrypted-file' }));
 app.use(express.json());
 
 app.use(validateUserMiddleware);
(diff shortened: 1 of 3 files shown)


MatissJanis:
Happy to review once conflicts are solved

tcrasset:
> Happy to review once conflicts are solved

Done :)

src/app-sync.js
```diff
@@ -12,7 +12,9 @@ import { SyncProtoBuf } from '@actual-app/crdt';
 
 const app = express();
 app.use(errorMiddleware);
+app.use(express.json());
 app.use(express.raw({ type: 'application/actual-sync' }));
+app.use(express.raw({ type: 'application/encrypted-file' }));
```
MatissJanis, line 17: Mind explaining this header? Why do we need it?

tcrasset: It's used on the `/upload-user-file` endpoint.


Source: https://github.com/actualbudget/actual/blob/d6afc85a8c179d0e4c6ac3249097a4d1989f9c60/packages/loot-core/src/server/cloud-storage.ts#L278

I alluded to it in one of my previous PRs, but the middlewares here are needed when running tests to make sure that the request/response bodies are parsed correctly, but on production they work without it. I'm not fully sure why, but with the middlewares, it works in the test, and outside.

MatissJanis (approved):

MatissJanis:
Thanks!

PR #427, opened 2024-08-16
fix payee name selection based on the amount

matt-fidd:
Fixes actualbudget/actual#3216

### DIFF
--- a/src/util/payee-name.js
+++ b/src/util/payee-name.js
@@ -5,7 +5,7 @@ function formatPayeeIban(iban) {
 }
 
 export const formatPayeeName = (trans) => {
-  const amount = trans.amount || trans.transactionAmount.amount;
+  const amount = trans.transactionAmount.amount;
   const nameParts = [];
 
   // get the correct name and account fields for the transaction amount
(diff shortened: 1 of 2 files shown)


src/util/payee-name.js
```diff
@@ -5,12 +5,13 @@ function formatPayeeIban(iban) {
 }
 
 export const formatPayeeName = (trans) => {
+  const amount = trans.amount || trans.transactionAmount.amount;
```
MatissJanis, line 8: **💬 suggestion:** ‏do we expect `trans.amount` to be `undefined`? If yes - then using `??` would be more appropriate.

```suggestion
  const amount = trans.amount ?? trans.transactionAmount.amount;
```

MatissJanis:
```javascript
console.log(0 ?? 12); // 0
console.log(0 || 12); // 12

console.log(undefined ?? 12); // 12
console.log(undefined || 12); // 12
```

matt-fidd: Fixed, turns out `trans.amount` is never set - I missed it when I ported the util over from actual.

MatissJanis (approved):

Soulplayer:
Bug remains. See https://github.com/actualbudget/actual/issues/3216#issuecomment-2299358407

PR #432, opened 2024-08-16
integrate FileService for app-sync.js

tcrasset:
Related to #420 

~Depends on #433, and on #425, and needs them to be merged first.~

Ready for review!


matt-fidd:
Thanks for this! I'm looking to have it fully reviewed this week, sorry it's taken a little while.
One of the tests is failing, please could you have a look?

tcrasset:
> Thanks for this! I'm looking to have it fully reviewed this week, sorry it's taken a little while. One of the tests is failing, please could you have a look?

Hey, thanks!
The test is flaky. Manual testing revealed that it's because we're not handling the database state correctly across the app. When all tests run in parallel across all the services, something goes wrong in the `accountDb`, and makes the test fail because the database couldn't be updated. Not only this test is affected, but other tests in `app-sync.test.js` as well. If I only run the `app-sync.test.js` tests, it never fails. It's only when they are run in parallel with the whole test suite. I haven't been able to pinpoint the exact problem. 


The current usage (one user per server) may make this a non-issue in production, but I think on a multi user system, with lots of calls, the database will be an issue.

matt-fidd:
Looking at this now

Just out of interest...

@coderabbitai review

MikesGlitch:
@coderabbitai review

src/app-sync.js
```diff
@@ -383,12 +346,31 @@ app.post('/delete-user-file', (req, res) => {
...
+    return;
   }
 
-  accountDb.mutate('UPDATE files SET deleted = TRUE WHERE id = ?', [fileId]);
+  filesService.update(fileId, new FileUpdate({ deleted: true }));
+
   res.send(OK_RESPONSE);
 });
+
+function boolToInt(deleted) {
```
matt-fidd, line 359: nit: could these function definitions be moved to the top please? `verifyFileExists` and `boolToInt`

src/app-sync.js
```diff
@@ -22,16 +32,11 @@ app.use(express.raw({ type: 'application/encrypted-file' }));
 app.use(express.json());
 
 app.use(validateUserMiddleware);
+app.use(errorMiddleware);
```
matt-fidd, line 35: This is already added in line 28

matt-fidd (changes requested):
Sorry this has taken so long to get a review in.
Two little nits.

matt-fidd (approved):
Thank you for this, massive improvement 🎉

Issue #431, opened 2024-08-16
[Bug]: GoCardless will implement strict ratelimiting


tcrasset:
### Verified issue does not already exist?

- [X] I have searched and found no existing issue

### What happened?

I got the following mail from GoCardless. 
I think we should grab those headers from each response and keep count of where we're at, so that we can:

1. return a comprehensible error to the user when GoCardless limits are reached
2. filtering unecessary sync events by the user from the actual app,to make sure that "unnecessary" calls are not being made to the GoCardless API, because 4 syncs a day is not a lot...

```
Dear Customer,

We are reaching out to inform you of some important updates regarding our API that will enhance your experience and provide greater visibility into bank API rate limits.

Improvements in the API
=======================

As part of our effort to enhance transparency around API rate limits, we are implementing additional information in our API response - rate limit headers.

What are Rate Limit Headers?
============================

Rate limit headers are response HTTP header fields included in API responses that will inform you about your current usage and remaining API call limits. These headers will offer you clearer insight into your rate limits for all API requests.

The new headers will provide key details such as the total request limit, the number of requests remaining within a specified timeframe, and the time until the limits reset.
The introduction of new API rate limit headers will include the following response headers:

General Rate Limits (for all API requests):
===========================================

    HTTP_X_RATELIMIT_LIMIT: Indicates the maximum number of allowed requests within the defined time window.
    HTTP_X_RATELIMIT_REMAINING: Shows the number of remaining requests you can make in the current time window.
    HTTP_X_RATELIMIT_RESET: Provides the time remaining in the current window.

Account Success Request (for account details, transactions, and balances):

    HTTP_X_RATELIMIT_ACCOUNT_SUCCESS_LIMIT: Indicates the maximum number of allowed requests within the defined time window.
    HTTP_X_RATELIMIT_ACCOUNT_SUCCESS_REMAINING: Shows the number of remaining requests you can make in the current time window.
    HTTP_X_RATELIMIT_ACCOUNT_SUCCESS_RESET: Provides the time remaining in the current window.

How are these changes being made?
=================================

Starting from 19 August 2024, we will implement a limit of 10 API requests per day for each access scope (details, balances, transactions) on account id level as part of our ongoing efforts to enhance operational effectiveness and minimize potential issues for all our clients. This phased approach will allow us to optimize our services progressively.

We plan to establish a limit of four requests per day for each access scope in the next phase.

[...]

PR #436, opened 2024-08-17
🧹 optimise GoCardless bank sync to use fewer api calls

matt-fidd:
Depends on merge of #435 

Part of #431 
Complimentary to PR in actual: actualbudget/actual#3279

The balance fetched from gocardless is used to calculate the starting balance, which is only relevant when a completely blank account is synced. This change lays the groundwork in the server to allow the client side to be more selective in what it retrieves.

### DIFF
--- a/src/app-gocardless/services/gocardless-service.js
+++ b/src/app-gocardless/services/gocardless-service.js
@@ -250,6 +250,69 @@ export const goCardlessService = {
 
   /**
    *
+   * @param requisitionId
+   * @param accountId
+   * @param startDate
+   * @param endDate
+   * @throws {AccountNotLinedToRequisition} Will throw an error if requisition not includes provided account id
+   * @throws {RequisitionNotLinked} Will throw an error if requisition is not in Linked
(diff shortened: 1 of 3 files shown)


src/app-gocardless/services/gocardless-service.js
```diff
@@ -244,6 +242,69 @@ export const goCardlessService = {
...
+      transactions.transactions?.pending,
+    );
+    const allTransactions = sortedBookedTransactions.map((t) => {
+      return { ...t, booked: true };
+    });
+    sortedPendingTransactions.forEach((t) =>
+      allTransactions.push({ ...t, booked: false }),
+    );
+    const sortedAllTransactions = bank.sortTransactions(allTransactions);
+
```
psybers, line 297: Can we refactor this out into a shared method to avoid the duplication with the other function?  That way there is less chance of divergence.

psybers (changes requested):

psybers (approved):

MikesGlitch (approved):

PR #439, opened 2024-08-20
fix gocardless error handling

matt-fidd:
Helps with #431 

It looks like GoCardless error handling wasn't working at all as `handleGoCardlessError` was never reached when client.${method} threw an error.

This changes fixes that, and allows the basic rate limit error to be sent back to the client.


src/app-gocardless/services/gocardless-service.js
```diff
@@ -35,24 +35,26 @@ const getGocardlessClient = () => {
...
-      throw new RateLimitError(response);
+      throw new RateLimitError(error);
     case 500:
-      throw new UnknownError(response);
+      throw new UnknownError(error);
     case 503:
-      throw new ServiceError(response);
+      throw new ServiceError(error);
     default:
       return;
```
tcrasset, line 59: Now that we explicitly pass in `error`, the `default` case should throw a generic error, not silently return

src/app-gocardless/services/tests/gocardless-service.spec.js
```diff
@@ -556,64 +470,56 @@ describe('goCardlessService', () => {
...
-    const response = { status_code: 503 };
+    const response = { response: { status: 503 } };
     expect(() => handleGoCardlessError(response)).toThrow(ServiceError);
   });
 
   it('does not throw an error for status code 200', () => {
-    const response = { status_code: 200 };
+    const response = { response: { status: 200 } };
     expect(() => handleGoCardlessError(response)).not.toThrow();
   });
```
tcrasset, line 524: This part is now changed, because we only pass in the response when it throws, not on a successful (200) call.

PR #442, opened 2024-08-30
Fix wrong payeeName used for KBC_KREDBEBB

ArnaudWeyts:
Fixes https://github.com/actualbudget/actual/issues/3303

Since we only have the `remittanceInformationUnstructured` available to us to extract the payee (for negative amounts), writing a dedicated integration for this seems reasonable.

~I couldn't really test this any further as I ran into GoCardless rate limits 🙈 But the basic implementation was working.~ Seems to work for my accounts now.

### DIFF
--- a/src/app-gocardless/bank-factory.js
+++ b/src/app-gocardless/bank-factory.js
@@ -2,6 +2,7 @@ import AbancaCaglesmm from './banks/abanca-caglesmm.js';
 import AmericanExpressAesudef1 from './banks/american-express-aesudef1.js';
 import BankinterBkbkesmm from './banks/bankinter-bkbkesmm.js';
 import Belfius from './banks/belfius_gkccbebb.js';
+import Berliner_Sparkasse_beladebexxx from './banks/berliner_sparkasse_beladebexxx.js';
 import BnpBeGebabebb from './banks/bnp-be-gebabebb.js';
 import DanskeBankDabNO22 from './banks/danskebank-dabno22.js';
 import EasybankBawaatww from './banks/easybank-bawaatww.js';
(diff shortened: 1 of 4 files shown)


ArnaudWeyts:
Did some final testing and had to do some minor fixes. The bank is working again for me 👌🏻

src/app-gocardless/bank-factory.js
```diff
@@ -21,6 +21,7 @@ import SpkMarburgBiedenkopfHeladef1mar from './banks/spk-marburg-biedenkopf-hela
 import SpkKarlsruhekarsde66 from './banks/spk-karlsruhe-karsde66.js';
 import VirginNrnbgb22 from './banks/virgin_nrnbgb22.js';
 import Berliner_Sparkasse_beladebexxx from './banks/berliner_sparkasse_beladebexxx.js';
+import KBCkredbebb from './banks/kbc_kredbebb.js';
```
matt-fidd, line 24: Could you move this to the correct place alpahetically please? Feel free to also move `Berliner_Sparkasse_beladebexxx` to the right place too.

src/app-gocardless/bank-factory.js
```diff
@@ -45,6 +46,7 @@ export const banks = [
   SpkKarlsruhekarsde66,
   VirginNrnbgb22,
   Berliner_Sparkasse_beladebexxx,
+  KBCkredbebb,
```
matt-fidd, line 49: Same with this one please, again - feel free to move `Berliner_Sparkasse_beladebexxx` at the same time.

matt-fidd (changes requested):
👋 Thanks for this! It looks good. Two tiny litte nits if you wouldn't mind fixing before I can merge.

src/app-gocardless/bank-factory.js
ArnaudWeyts: Done ✅

matt-fidd (approved):
Thank you!

PR #445, opened 2024-09-03
Fix BancSabadell payee name based on the transaction amount

davidmartos96:
The goCardless API for this bank doesn't return creditor and debtor names with its appropriate role based on the transaction amount, which makes Actual infer an incorrect payee name. So we just use the remittance information directly.

Hopefully this is ok, first time contributing.
Great project!


MatissJanis (approved):
Thanks

PR #448, opened 2024-09-06
Fix pending transaction amount sign and payee name for National Bank of Greece import

mezger6:
The pending purchases retrieved from National Bank of Greece are erroneously sent as positive, and have a static unnecessary prefix ("ΑΓΟΡΑ") added to the remittance information, causing them to be added as deposits, and preventing  the eventual booked transactions from being properly matched.

This fix adds the minus sign and removes the prefix for the problematic pending transactions.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new bank entity, NbgEthngraaxxx, expanding the list of available banks.
	- Added robust data normalization and balance calculation methods for the NBG_ETHNGRAAXXX bank integration.

- **Bug Fixes**
	- Resolved issues with pending purchases amount sign and payee name during the import process, enhancing financial data accuracy.

- **Tests**
	- Implemented a test suite for the NbgEthngraaxxx module to validate transaction normalization logic.


MatissJanis (approved):
Thanks!

PR #450, opened 2024-09-09
[WIP] Fix wrong payeeName used for CBC_CREGBEBB

MMichotte:
Fixes [actualbudget/actual#3216](https://github.com/actualbudget/actual/issues/3216) (Complementary)
=> Incorrect payeeName for both KBC and CBC banks.

A first fix was implemented to support KBC bank but not for CBC (see: #442). 
This PR implements the same fallback logic as for KBC but for CBC bank (CBC_CREGBEBB).


MMichotte:
Rebased from root by mistake... Will simply recreate PR!

PR #452, opened 2024-09-11
Update bank-factory.js

molnarti:
added REVOLUT_REVOLT21 & VUB_BANKA_SUBASKBX to banks with limited history


MatissJanis (approved):
Thanks

PR #463, opened 2024-09-23
[WIP] Add "Boursorama Bank" (FR) to list of banks

bakoff44:
Importing datas with default integration, will insert unwanted datas in operation names.
This PR is done to ignore theses data when imported from GoCardless


matt-fidd (commented):
Thanks for this! Would it be possible for you to add some test cases for these changes in as well please?

PR #470, opened 2024-10-01
feat: sort transactions better

UnderKoen:
I noticed that the transaction that were synced were really out of order of how they happend. This should fix this by also sorting  on time


matt-fidd (approved):
Thank you for this, I like the idea. I think there is a consideration to be made regarding the performance hit, but I did some testing and it doesn't seem to cause a pronounced slowdown even with a full 90 days of data.

PR #474, opened 2024-10-05
fix: Sabadell Bank regression, missing date field during normalization

davidmartos96:
Fixes a regression from a PR of mine from some weeks ago #445. I forgot to include the `date` field in the normalized transaction, which fails when syncing.


MikesGlitch (approved):
Looks great! Thanks

PR #478, opened 2024-10-13
refactor: set correct log level

rare-magma:
This PR will set the correct log level for bank integration messages which previously were logging unnecessary details at info level.


matt-fidd (approved):
Thanks for this!

PR #481, opened 2024-10-17
Add Fineco to banks with limited history

ftruzzi:
Hi, and thank you for your work!

This PR adds Fineco bank to the list of bank with limited history.

While I'm not actually sure that the bank actually has limited history (it probably doesn't), this is the only way I've been able to consistently sync the bank to Actual. Otherwise, I would either get a `409 Conflict` from GoCardless with the Actual UI indefinitely waiting for a sync response, or the Actual UI would crash / go white when trying to add the account. Furthermore, every failed sync attempt uses up the available requisitions on GoCardless which are not reset until the following month.

From what I understand, users usually start their budgeting journey on the current/previous month, so hopefully this will be a welcome change.

Thanks!


matt-fidd:
Thank you for this!

PR #482, opened 2024-10-22
Do not request transactions when listing accounts

psybers:
This utilizes the new support for `balances-only=1` in the SimpleFIN API so that when requesting a list of accounts (to link to) we only grab the account data, and do not pull the transactions.


psybers:
See: https://github.com/simplefin/simplefin.github.com/pull/26 for details on the new API being used.

psybers:
I verified listing accounts on this branch still works (for my data).

matt-fidd (approved):
LGTM, can't test but the logic looks sound

PR #485, opened 2024-10-26
Check SimpleFIN access key format

psybers:
Adds a check to ensure the SimpleFIN accessKey is in the expected format.  Helps avoid silent errors if the key was invalid (e.g., if CloudFlare blocked the call).


matt-fidd (approved):
Looks to work with the SimpleFin test account

PR #501, opened 2024-11-12
Add "Eurobank Ergasias" bank (GR) to the list of banks with limited h…

giliopoulos:
Add "Eurobank Ergasias" bank (GR) to the list of banks with limited history


twk3:
Thanks @giliopoulos . Just to double check, the problem you are running into with this bank is the same as https://bankaccountdata.zendesk.com/hc/en-gb/articles/11529718632476-Extended-history-and-continuous-access-edge-cases were you don't get ongoing new data in bank sync because the request was for over 90 days.

giliopoulos:
Actually i am able to get new transactions when i sync with the bank from day to day.. When i did my first sync i got the transactions for the last 90 days. So looking through the documentation, i thought this way i could extend the period for the fetched transactions to over 90 days. 

Sorry for the misunderstanding.

PR #502, opened 2024-11-13
use built-in Docker init, remove tini

meonkeys:
We no longer need tini with recent versions of Docker.

See:

* https://docs.docker.com/reference/cli/docker/container/run/#init
* https://github.com/krallin/tini?#using-tini

Tested and seems to work as expected when `--init` is included in `docker run` command line or `init: true` is included in the Docker Compose service stanza.

I **think** it's OK to exclude the `-g` argument (pass along signals to process group) since we only run the one node process, but I'm not certain.

Look for a correspoding change in the docs repo.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,6 @@ COPY yarn.lock package.json .yarnrc.yml ./
 RUN yarn workspaces focus --all --production
 
 FROM node:18-bookworm-slim as prod
-RUN apt-get update && apt-get install tini && apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
 ARG USERNAME=actual
 ARG USER_UID=1001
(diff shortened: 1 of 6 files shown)


meonkeys:
see https://github.com/actualbudget/docs/pull/486

meonkeys:
From a Discord chat discussion: if people upgrade Docker containers and get this change, their containers will shut down more slowly (10sec by default, IIRC) until/unless `--init` is added to the `docker run` command line or `init: true` is included in their service stanza in their Docker Compose config file.

billneftleberg:
Do we still need tini with recent Docker versions if --init or init: true works, and is it safe to skip the -g argument?

meonkeys:
Based on [Daniel's message](https://discord.com/channels/937901803608096828/1033401142601859112/1306296960226955385) I think we're OK to merge? I'm fine either way

meonkeys:
Hmm, can't say I agree 100% with the AI, but I'll make one of the suggested changes.

( init isn't _required_ any more than tini was, but it's nice when containers shut down quickly and cleanly )

meonkeys:
@coderabbitai resolve

matt-fidd:
I think for consistency it's best if we keep tini, removing it may break some people's configuration as they're not using init

PR #503, opened 2024-11-14
remove axios dependency

matt-fidd:
Reported in discord, this is causing an issue because it's not in the package.json

To avoid adding a large dependency I've rolled our own here, referencing https://github.com/axios/axios/blob/5d54d22321b3c3d29c5dff5e8086cd9db7e2d101/lib/helpers/isAxiosError.js#L12

### DIFF
--- a/src/app-gocardless/app-gocardless.js
+++ b/src/app-gocardless/app-gocardless.js
@@ -1,4 +1,3 @@
-import { isAxiosError } from 'axios';
 import express from 'express';
 import path from 'path';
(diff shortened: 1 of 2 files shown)


MikesGlitch (approved):
Not sure why we need to check if it's an axios error - I'm assuming that'll never be the case here since we don't use axios. 

However, this will definately fix the error reported so happy to go with this.

matt-fidd:
Closing, root cause of the error is not what we thought

PR #504, opened 2024-11-19
fix bug in batch SimpleFIN startDate logic

matt-fidd:
The comparison here wasn't like for like and was always failing, resulting in the start date being ignored

### DIFF
--- a/src/app-simplefin/app-simplefin.js
+++ b/src/app-simplefin/app-simplefin.js
@@ -210,12 +210,13 @@ function getAccountResponse(results, accountId, startDate) {
       dateToUse = trans.posted;
     }
 
-    newTrans.bookingDate = getDate(new Date(dateToUse * 1000));
-    if (newTrans.bookingDate < startDate) {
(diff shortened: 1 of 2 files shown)


matt-fidd:
@coderabbitai full review

It seems to be struggling to pass...

youngcw (approved):

src/app-simplefin/app-simplefin.js
```diff
@@ -210,12 +210,13 @@ function getAccountResponse(results, accountId, startDate) {
       dateToUse = trans.posted;
     }
 
-    newTrans.bookingDate = getDate(new Date(dateToUse * 1000));
```
psybers, line 213: Does SF not use `bookingDate`?

matt-fidd: No, all bank sync looks to use `date` https://github.com/actualbudget/actual/blob/c626fc2f17199a327c8dcd80e93c3549e8926240/packages/loot-core/src/server/accounts/sync.ts#L326

PR #511, opened 2024-11-28
Fix Hype Bank sync (`HYPE_HYEEIT22`) 

### LINKED ISSUE #462: [Bug]: Extended history of special banks sync with gocardless

### What happened?

I am aware that some banks ([listed here](https://bankaccountdata.zendesk.com/hc/en-gb/articles/11529718632476-Extended-history-and-continuous-access-edge-cases)) have a limitations on the number of days transactions are available continuously.

This should be addressed by actual in this PR https://github.com/actualbudget/actual-server/pull/355 and similar ones.

However, when using HYPE_HYEEIT2 (Hype - Italy), even if transactions are fetched the very first time correctly, only the same old transactions are fetched afterwards. In fact, any new transaction is not synced even with manual bank sync.

I verified that HYPE_HYEEIT2 is included in the BANKS_WITH_LIMITED_HISTORY array [here](https://github.com/hostyn/actual-server/blob/7b314e3b250404546e9c4f4d5eb3b4c949356100/src/app-gocardless/bank-factory.js#L94) 

_No response_

In a kubernetes cluster

Firefox, Chrome, Safari

### Operating System

Mac OSX

guglicap:
Hello, as title says, Hype Bank sync is currently broken.
This is because this bank is one of the limited history ones, but the wrong institution ID is set in the `BANKS_WITH_LIMITED_HISTORY` array.
This fixes #462  by replacing the wrong id with the correct one.


matt-fidd (approved):
Thanks for this!

PR #515, opened 2024-12-05
[WIP] 🔖 (24.12.0)

matt-fidd:


MikesGlitch (approved):

jfdoming (changes requested):
Just so we don't accidentally merge, I think we still need to bump the web package here

matt-fidd:
> Just so we don't accidentally merge, I think we still need to bump the web package here 

Yep, I'll do it once I've published to NPM

PR #523, opened 2024-12-13
🐛 Fix for Electron OpenID Authentication

lelemm:
Since Electron window can't go to another URL. I had to change the way it logins in for OpenID. To make this work, I had to change a validation of the redirect uri


src/accounts/openid.js
```diff
@@ -304,7 +304,10 @@ export function isValidRedirectUrl(url) {
     const redirectUrl = new URL(url);
     const serverUrl = new URL(serverHostname);
 
-    if (redirectUrl.hostname === serverUrl.hostname) {
+    if (
+      redirectUrl.hostname === serverUrl.hostname ||
+      redirectUrl.hostname === 'localhost'
```
MikesGlitch, line 309: IMO this should be configurable. I'd like to be able to specify how I want to lock down my server in the config. 

Maybe by default we allow localhost and the servers url, but allowing us to change this would be very useful.

E.g. I'm testing against your Web PR and I want to be able to allow the redirectURL of that PR environment. 

There's also the case where some people use app.actualbudget.org or edge.actualbudget.org with their own servers. I think I'd like to be able to test my server against those deployed environments.

lelemm: do you mean put a env/config variable to not allow localhost?

MikesGlitch: So It looks like the server hostname and localhost are always required due to the way electron and the server client are hosted, so that part's good to go and fixes the issue.

I'm thinking it would be good to have env variables/config for specifying our own redirectUrl's. If I want to use a client hosted at a different URL from the serverurl I need to be able to specify it as an allowed redirectUrl. This would allow us to test the latest edge code hosted at edge.actualbudget.org against our servers. 

It's a nice to have I guess. It's probably an overreach for this PR - I can add it in another PR.

lelemm: got it. i can make another PR to this after if you want

MikesGlitch (approved):

PR #533, opened 2025-01-03
Add default undefined option for payeename in KBC integration

robxgd:
Added a default undefined option as Payeename for the KBC integration as this is not always available as described in issue:  #532 .


MatissJanis (approved):
LGTM

PR #534, opened 2025-01-03
Improve abn amro sync

UnderKoen:
For payment made with google pay this also cleanes the payee_name.

Removes the removal of certain terms in the notes. As for me it really helpful to have BEA and GEA in the notes. (GEA is for atm withdraws) 

The cleaning should be achieved with https://github.com/actualbudget/actual/issues/3606 I think. @nsulzer would you be opposed by this?

### DIFF
--- a/src/app-gocardless/banks/abnamro_abnanl2a.js
+++ b/src/app-gocardless/banks/abnamro_abnanl2a.js
@@ -37,13 +37,6 @@ export default {
     transaction.debtorName = transaction.debtorName || payeeName;
     transaction.creditorName = transaction.creditorName || payeeName;
 
-    // There are anumber of superfluous keywords in the remittanceInformation.
-    // Remove them to aboid clutter in notes.
-    const keywordsToRemove = ['.EA, Betaalpas', ',PAS\\d{3}', 'NR:.+, '];
-    const regex = new RegExp(keywordsToRemove.join('|'), 'g');
-    transaction.remittanceInformationUnstructured =
-      remittanceInformationUnstructured.replace(regex, '').trim();
-
     return {
       ...transaction,
       payeeName: formatPayeeName(transaction),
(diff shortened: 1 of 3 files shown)


nsulzer:
Hiya @UnderKoen, after tinkering with the rule action templating a bit, I have no objections. I agree that it's convinent to keep all information and let people decide for themselves what to keep. Also, thanks for cleaning up the payeeName extraction 😅.

thomwiggers:
The way the approach in the PR works is that it relies on particular payment service providers that inject their own name into the transaction description; e.g., in the provided test case, this is CCV and the regex relies on `CCV*` being part of the string to strip it. 

This is not a very solid approach; many payment terminals *do not* have the same treatment. E.g., looking at my own transactions, Albert Heijn is listed as something like `BEA, Apple Pay Albert Heijn,PAS123,NR:123456, 01.01.24/10:59, ZAANDAM`. If the intention of this PR was to remove Google Pay, it seems better to explicitly include `((Google|Apple) Pay)?` into the regex.

thomwiggers:
Or were you intending to *just* remove the payment service provider from the vendor name? (because AFAIK that's not related to Google Pay at all).

thomwiggers:
Never mind, I misunderstood how the parsing changed: it now only matches on line 2 of the input and that automatically removes the Apple Pay indicator as well.

UnderKoen:
Hey @thomwiggers , 
This approach uses the fact that the payee name is on a seprate line. Most of the times the second one.

So for your example in #540 
```
remittanceInformationUnstructuredArray: [
	'BEA, Apple Pay',
	'Other payee name,PAS123', <-- this one
	'NR:123A4B, 09.12.23/15:43',
	'CITY',
],
```

It would look for the line with `PAS...` in it

thomwiggers:
Yeah I clearly wasn't awake enough last night, all good!

PR #535, opened 2025-01-05
Corner case transaction for GoCardLess adapater (ING Romania)

spideraxal:
Hello,

I am updating my adapter for ING Romania. It seems there is a special transaction type (insurance provided by the bank itself for example), which does not have the `remittanceInformationUnstructured` field, only the `proprietaryBankTransactionCode` one.

This led to a `Cannot read properties of undefined (reading 'toLowerCase')` error, so no new transactions were synced. This commit fixes the issue.

image

### DIFF
--- a/src/app-gocardless/banks/ing-ingbrobu.js
+++ b/src/app-gocardless/banks/ing-ingbrobu.js
@@ -16,7 +16,7 @@ export default {
     //For deduplication to work better, payeeName needs to be standardized
     //and converted from a pending transaction form ("payeeName":"Card no: xxxxxxxxxxxx1111"') to a booked transaction form ("payeeName":"Card no: Xxxx Xxxx Xxxx 1111")
     if (transaction.transactionId === 'NOTPROVIDED') {
-      //Some corner case transactions have a special field called `proprietaryBankTransactionCode`, this need to be copied to `remittanceInformationUnstructured`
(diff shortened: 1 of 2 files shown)


src/app-gocardless/banks/ing-ingbrobu.js
```diff
@@ -18,7 +18,17 @@ export default {
...
         transaction.transactionId = transaction.internalTransactionId;
+
+        //Some corner case transactions have a special field called `proprietaryBankTransactionCode`, this need to be copied to `remittanceInformationUnstructured`
+        if (
+          transaction.proprietaryBankTransactionCode &&
+          !transaction.remittanceInformationUnstructured
+        ) {
+          transaction.remittanceInformationUnstructured =
+            transaction.proprietaryBankTransactionCode;
+        }
```
UnderKoen, line 29: As you repeat this on line 45 could you move this to line 18? to prevent duplicate code

UnderKoen (changes requested):
Small maintance nitpick, otherwise approved

src/app-gocardless/banks/ing-ingbrobu.js
spideraxal: Done!

UnderKoen (approved):

PR #537, opened 2025-01-05
Add GoCardless integration for COMMERZBANK_COBADEFF

nsulzer:
Add GoCardless integration for COMMERZBANK_COBADEFF, with a bunch of cleanup for remittance information that will become the notes. The source data for this bank is a bit messy to format unfortunately.

Question to the reviewer: Is the implementation of the optional `iban` property for `creditorAccount` correct/clean? Not a dev, sorry.

### DIFF
--- a/src/app-gocardless/banks/commerzbank_cobadeff.js
+++ b/src/app-gocardless/banks/commerzbank_cobadeff.js
@@ -7,8 +7,6 @@ export default {
 
   institutionIds: ['COMMERZBANK_COBADEFF'],
 
-  accessValidForDays: 179,
-
   normalizeTransaction(transaction, _booked) {
     // remittanceInformationUnstructured is limited to 140 chars thus ...
     // ... missing information form remittanceInformationUnstructuredArray ...
--- a/src/app-gocardless/banks/commerzbank_cobadeff.js
+++ b/src/app-gocardless/banks/commerzbank_cobadeff.js
@@ -1,5 +1,4 @@
 import Fallback from './integration-bank.js';
-import { amountToInteger, printIban } from '../utils.js';
 import { formatPayeeName } from '../../util/payee-name.js';
 
 /** @type {import('./bank.interface.js').IBank} */
@@ -9,18 +8,6 @@ export default {
   institutionIds: ['COMMERZBANK_COBADEFF'],
 
   accessValidForDays: 179,
-
-  normalizeAccount(account) {
-    return {
-      account_id: account.id,
-      institution: account.institution,
-      mask: account.iban.slice(-4),
-      iban: account.iban,
-      name: [account.name, printIban(account)].join(' '),
-      official_name: account.product,
-      type: 'checking',
-    };
-  },
 
   normalizeTransaction(transaction, _booked) {
     // remittanceInformationUnstructured is limited to 140 chars thus ...
@@ -43,6 +30,7 @@ export default {
     keywords.forEach((keyword) => {
       transaction.remittanceInformationUnstructured =
         transaction.remittanceInformationUnstructured.replace(
+          // There can be spaces in keywords
           RegExp(keyword.split('').join('\\s*'), 'gi'),
           ', ' + keyword + ' ',
         );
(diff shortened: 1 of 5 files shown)


src/app-gocardless/banks/commerzbank_cobadeff.js
```diff
@@ -0,0 +1,87 @@
...
+    return {
+      account_id: account.id,
+      institution: account.institution,
+      mask: account.iban.slice(-4),
+      iban: account.iban,
+      name: [account.name, printIban(account)].join(' '),
+      official_name: account.product,
+      type: 'checking',
+    };
+  },
```
UnderKoen, line 23: As this is the same as in the [fallback](https://github.com/actualbudget/actual-server/blob/master/src/app-gocardless/banks/integration-bank.js) you could remove this and rely on the fallback

```suggestion
```

src/app-gocardless/banks/commerzbank_cobadeff.js
```diff
@@ -0,0 +1,87 @@
...
+      'Mandatsref:',
+      'Gläubiger-ID:',
+      'SEPA-BASISLASTSCHRIFT',
+      'Kartenzahlung',
+      'Dauerauftrag',
+    ];
+    keywords.forEach((keyword) => {
+      transaction.remittanceInformationUnstructured =
+        transaction.remittanceInformationUnstructured.replace(
+          RegExp(keyword.split('').join('\\s*'), 'gi'),
```
UnderKoen, line 46:
```suggestion
					// There can be spaces inbetween keywords
          RegExp(keyword.split('').join('\\s*'), 'gi'),
```

src/app-gocardless/banks/commerzbank_cobadeff.js
```diff
@@ -0,0 +1,87 @@
...
+   */
+  calculateStartingBalance(sortedTransactions = [], balances = []) {
+    const currentBalance = balances.find(
+      (balance) => 'expected' === balance.balanceType,
+    );
+
+    return sortedTransactions.reduce((total, trans) => {
+      return total - amountToInteger(trans.transactionAmount.amount);
+    }, amountToInteger(currentBalance.balanceAmount.amount));
+  },
```
UnderKoen, line 86: I think this could also just use the fallback?

UnderKoen (changes requested):

UnderKoen:
> Question to the reviewer: Is the implementation of the optional iban property for creditorAccount correct/clean? Not a dev, sorry.

Seems fine 👍

src/app-gocardless/gocardless-node.types.ts
```diff
@@ -356,7 +356,11 @@ export type Transaction = {
   /**
    * Account reference, conditional
    */
-  creditorAccount?: string;
+  creditorAccount?:
+    | string
+    | {
+        iban?: string;
+      };
```
matt-fidd, line 363: As it happens, this can just be an object. It was wrongly typed before.

```suggestion
  creditorAccount?: {
    iban: string;
  };
```

nsulzer: @matt-fidd I did this, but it seems to fail to build, as there are cases/banks where `creditorAccount` really is just a string. Shall we just keep it as the original optional string or object?

matt-fidd: Ah okay, leave it as you've done it for now and we can fix in other places in another PR

nsulzer: Alright :) Thanks for the quick feedback.

src/app-gocardless/banks/commerzbank_cobadeff.js
nsulzer: Yup, removed, thanks.

src/app-gocardless/banks/commerzbank_cobadeff.js
```diff
@@ -0,0 +1,56 @@
...
+import { formatPayeeName } from '../../util/payee-name.js';
+
+/** @type {import('./bank.interface.js').IBank} */
+export default {
+  ...Fallback,
+
+  institutionIds: ['COMMERZBANK_COBADEFF'],
+
+  accessValidForDays: 179,
+
```
UnderKoen, line 11: With https://github.com/actualbudget/actual-server/pull/551 in mind this property can be removed.

```suggestion
```

UnderKoen (approved):

PR #538, opened 2025-01-06
Fix WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match

lnagel:
Further info https://docs.docker.com/reference/build-checks/from-as-casing/


matt-fidd (commented):
Thank you for this!

matt-fidd (approved):

PR #557, opened 2025-01-27
dynamically load GoCardless handlers

matt-fidd:


UnderKoen (approved):
Awesome! 

Would it now be possible to define a custom handler with a docker volume?

matt-fidd:
> Awesome!
> 
> Would it now be possible to define a custom handler with a docker volume?

Hmm.. perhaps. I don't know too much about docker.

If I understand correctly we might be able to define a specific directory that it would look in for "extras" and mount that as a volume

matt-fidd:
I'll merge this after the freeze since it doesn't actually do anything different to master

